PR #3, opened 2016-09-01
arm: Add PSR helper type

lioncash:
Should be helpful later on when other modes and such are supported.

At the moment I've just made it proxy through existing functions in the LocationDescriptor class since I wasn't sure what the best way of doing that would be while maintaining the return-by-value interface for LocationDescriptor instances.

### DIFF
--- a/src/frontend/arm_types.h
+++ b/src/frontend/arm_types.h
@@ -13,6 +13,7 @@
 #include "common/assert.h"
 #include "common/common_types.h"
 #include "frontend/arm/FPSCR.h"
+#include "frontend/arm/PSR.h"
 
 namespace Dynarmic {
 namespace Arm {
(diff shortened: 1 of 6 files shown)


src/frontend/arm_types.h
```diff
@@ -73,57 +74,64 @@ enum class SignExtendRotation {
...
-    bool EFlag() const { return eflag; }
+    bool TFlag() const { return cpsr.T(); }
+    bool EFlag() const { return cpsr.E(); }
+
+    Arm::PSR CPSR() const { return cpsr; }
     Arm::FPSCR FPSCR() const { return fpscr; }
 
     bool operator == (const LocationDescriptor& o) const {
-        return std::tie(arm_pc, tflag, eflag, fpscr) == std::tie(o.arm_pc, o.tflag, o.eflag, o.fpscr);
+        return std::tie(arm_pc, cpsr, fpscr) == std::tie(o.arm_pc, o.cpsr, o.fpscr);
```
MerryMage: Given that LocationDescriptor is used as a key for identifying emitted basicblocks, this would result in quite a few redundant basic blocks. Perhaps mask psr as is done with fpscr?

lioncash: What is the FPSCR mask supposed to represent?

MerryMage: "bits that compilation depends on"

MerryMage:
LGTM

PR #7, opened 2016-09-02
load_store: Add correctness checks for STRD variants

lioncash:
Also, STRD doesn't allow the use of the PC in either Rt or Rt2


MerryMage:
LGTM. Some unpredictable behaviour still remains in the remaining stores (on todo-list).

PR #11, opened 2016-09-02
load_store: Add checks for unpredictability to other singular store instructions

lioncash:
Gets the single item stores out of the way.


MerryMage:
Looks good.

PR #20, opened 2016-09-04
arm_types: Split out LocationDescriptor

lioncash:
This isn't really an ARM-specific type, since it's used to indicate a Block location.

This also prevents inclusion of ARM-specifc types when a LocationDescriptor is the only thing needed.


MerryMage:
LGTM

PR #25, opened 2016-09-05
arm_types: Move into arm folder

lioncash:
Makes the base directory nice and empty except for directories (and keeps all the arm stuff together)


MerryMage:
LGTM

PR #28, opened 2016-09-07
memory_pool: Deduplicate slab allocation code

lioncash:


src/common/memory_pool.cpp
```diff
@@ -28,17 +26,23 @@ Pool::~Pool() {
...
-    void* ret = (void*)current_ptr;
+    void* ret = static_cast<void*>(current_ptr);
     current_ptr += object_size;
     remaining--;
 
     return ret;
 }
 
+void Pool::AllocateNewSlab()
+{
```
MerryMage: minor: Codebase uses attached braces (w/o exceptions I think).

lioncash: Whoops, updated.

PR #31, opened 2016-09-11
[rfc] fuzz_arm: New ARM tests

MerryMage:
One thing I don't like about doing it this way is it duplicates validity checking across two places.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -3,6 +3,7 @@ set(SRCS
     arm/fuzz_thumb.cpp
     arm/test_arm_disassembler.cpp
     arm/test_thumb_instructions.cpp
+    arm/valid_arm.cpp
     main.cpp
     rand_int.h
     skyeye_interpreter/dyncom/arm_dyncom_dec.cpp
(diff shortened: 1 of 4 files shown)


MerryMage:
Comment: Should probably move this validity checker into the main codebase; check validity before translating instructions and remove the asserts that currently exist.

This would mean that this could also be reused in the disassembler if this is desirable.

lioncash:
What other place is validity checking done at (just to make sure I'm on the same page)?

MerryMage:
@lioncash

Currently just in the translator ([I am referring to the unpredictable instruction checks](https://github.com/MerryMage/dynarmic/blob/master/src/frontend/translate/translate_arm/data_processing.cpp#L119-L120).)

lioncash:
Hmm, I guess if all the unpredictable checks come before the actual instruction functionality (and that assumption holds for all instructions), then perhaps a checker of some sort can be made that just does the checks before executing the instruction, so you'd have it in between like so:

```
[Instruction Decode] → [Unpredictable Checker] → [Instruction Translation]
```

which would also let them be used for other purposes.

MerryMage:
Will do this eventually™.

PR #32, opened 2016-09-14
disassembler: Provide includes to function declarations

lioncash:
Otherwise missing declaration warnings can happen.


MerryMage:
LGTM

PR #33, opened 2016-09-16
decoder: Generify the matcher interface

lioncash:
Gets rid of a bit of duplication while remaining compatible with the current interfaces in place.


MerryMage:
I like this.

LGTM

PR #36, opened 2016-11-08
Fixed compilation issues with GCC 4.9.3.

spixi:
This fixes compilation issues with GCC 4.9.3.

The C++ standard does not allow pointers to references and parameters in constexpr functions, which causes several errors.


lioncash:
> The C++ standard does not allow pointers to references and parameters in constexpr functions, which causes several errors.

This is incorrect and demonstrably false.

https://godbolt.org/g/y8dgBh

spixi:
So your code gives me a ```<Compilation failed>```

spixi:
So, when I add the constexpr I got the following issue:

```
In file included from /home/mari/citra/externals/dynarmic/src/frontend/arm/types.cpp:10:0:
/home/mari/citra/externals/dynarmic/src/./common/bit_util.h: In instantiation of 'constexpr bool Dynarmic::Common::Bit(size_t, T) [with T = short unsigned int; size_t = long unsigned int]':
/home/mari/citra/externals/dynarmic/src/frontend/arm/types.cpp:44:36:   required from here
/home/mari/citra/externals/dynarmic/src/./common/bit_util.h:57:1: error: body of constexpr function 'constexpr bool Dynarmic::Common::Bit(size_t, T) [with T = short unsigned int; size_t = long unsigned int]' not a return-statement
 }
 ^
```


Then the second issue is

```
In file included from /home/mari/citra/externals/dynarmic/src/backend_x64/abi.cpp:21:0:
/home/mari/citra/externals/dynarmic/src/./common/iterator_util.h: In instantiation of 'auto Dynarmic::Common::detail::ReverseAdapter<T>::begin() [with T = const std::array<Dynarmic::BackendX64::HostLoc, 6ul>&]':
/home/mari/citra/externals/dynarmic/src/backend_x64/abi.cpp:107:5:   required from 'void Dynarmic::BackendX64::ABI_PopRegistersAndAdjustStack(Xbyak::CodeGenerator*, size_t, const RegisterArrayT&) [with RegisterArrayT = std::array<Dynarmic::BackendX64::HostLoc, 6ul>; size_t = long unsigned int]'
/home/mari/citra/externals/dynarmic/src/backend_x64/abi.cpp:119:73:   required from here
/home/mari/citra/externals/dynarmic/src/./common/iterator_util.h:22:31: error: 'rbegin' was not declared in this scope
         return rbegin(iterable);
                               ^
```

lioncash:
> So, when I add the constexpr I got the following issue:

If it works in Godbolt's compiler explorer on 4.9.3, then that's an issue with your particular installation.

> Then the second issue is

We use features of C++ that are only supported in GCC5 and up (and it will likely stay that way, since GCC has better support for standardized type traits (namely the is_trivial family among other things).

> and when I add #include <vector> in src/commons/iterator_util.h and replace rbegin and rend with vector<T>::rbegin and vector<T>::rend I get:

That issue is because std::vector::rbegin and std::vector::rend are incorrect to use there. It's not a generic function like non-member std::rbegin and std::rend are. You're using it in the context where a std::array would be used, which is incompatible with std::vector. The error message even indicates this.

spixi:
I did not know  ```std::rbegin``` and ```std::rend```. I googled them and realized that they are C++14 functions. I compiled dyrnamic with C++11. That's the reason, why it did not work. Sorry for the circumstances.

spixi:
@lioncash Nevertheless, the compiling result for the constexpr and the non-constexpr version is the same if you use at least ```-O1```.

MerryMage:
Closed for reasons above.

spixi:
Even when I compile with ```-std=c++14 -w``` compilation still fails.

PR #38, opened 2016-11-10
Fixed compilation issues with GCC 4.9.3.

spixi:
The compilation issue still occurs, even if I compile with ```-std=c++14```. This will fix it.


spixi:
Also note that

```
/home/mari/citra/externals/dynarmic/src/./backend_x64/jitstate.h:24:29: error: missing initializer for member 'std::array<unsigned int, 16ul>::_M_elems' [-Werror=missing-field-initializers]
     std::array<u32, 16> Reg{}; // Current register file.
                             ^
```
if you don't use compile dynarmic with ```-w```

lioncash:
From #36 

> Even when I compile with -std=c++14 -w compilation still fails. Your reverse iterator wrapper is only used on one place and the constexpr functions are not really constexpr functions because they depend on runtime variables. Please reopen this issue. Thank you.

They most certainly are constexpr functions. Whether they are used mostly in the context of runtime variables does not change that. `constexpr` literally means it's a function that is allowed to be evaluated at compile-time if it's capable of doing so in a given context. Whether or not it's mostly used with runtime variables does not change that.

Again, if your installation of GCC 4.9.3 cannot compile this usage of constexpr successfully, yet the 4.9.3 on Godbolt's compiler explorer can, this is literally a problem with your installation. Please do not remove the specifiers.

I also completely disagree with rolling back to support for 4.9.3, because that effectively cripples what can be done in terms of using the standard library.

spixi:
I try compiling 5.3.0 for about three days. 4.9.3 works fine. If you see on Godbolt it makes no difference on the generated machine code whether you add the ```constexpr``` keyword or not, if you at least use ```-O3```. I also don't understand the requirement of a reverse iterator. This is already possible with much less coding (and actually faster) as shown in my example.

lioncash:
> I try compiling 5.3.0 for about three days, always getting segmentation faults, only for this. 4.9.3 works fine. 

Does your distro not provide any recent compiler versions out of the box? This really isn't our problem to solve, especially if it's not our code that's segfaulting.

> If you see on Godbolt it makes no difference on the generated machine code whether you add the constexpr keyword or not, if you at least use -O1. 

The point of constexpr is it also allows functions and variables to be used in constant expressions, whatever the compiler produces is just an additional benefit. Again, please do not remove the specifiers. We can go in circles about this and I'll keep saying the same thing.

> (and actually faster...

No it isn't; compilers will generally elide this sort of thing: [example](https://godbolt.org/g/fo5vk5). The point of `Reverse` is to provide shorthand for reverse iteration across a sequence, similar to a basic range-based for where explicit specifying of the begin and end isn't necessary.

spixi:
Finally got gcc-5.30 compiled after I increased my swap file to 24GB. I'll try it tomorrow.

PR #43, opened 2016-11-24
libfmt: Update version to current master / CMake: Add option to disable use of bundled fmtlib

yuriks:
The CMake thing is kinda gross, but it's simple. Planning on using fmt in citra and the two copies of the targets conflict.


MerryMage:
Perhaps add `option(DYNARMIC_USE_BUNDLED_FMT "Use bundled fmtlib" ON)` to the root CMakeFiles.txt instead.

yuriks:
@MerryMage I had that at first, but it means that a cached value overrides any value set by the main script. So you have to use "set(DYNAMIRC_USE_BUNDLED_FMT OFF CACHED "" FORCE)" anyway, and that that point you might as well just use a regular variable, since I don't think there's any reason to not use the bundled fmt if you're running the dynarmic CMakeLists directly.

MerryMage:
Merged

PR #45, opened 2016-11-25
Implemented the ARM UHADD8 instruction.

Subv:
The x64 implementation uses the SSSE3 instruction PSHUFB.
A non-SSE fallback is provided in case the CPU doesn't support it.


MerryMage:
LGTM

yuriks:
Isn't the fallback integer code faster than the SSE version...?

PR #46, opened 2016-11-25
mp: Fix static_assert condition

lioncash:
Not an issue currently, but this would have prevented type inspection on the last function parameter.


MerryMage:
LGTM

PR #47, opened 2016-11-26
Implement SHADD8 and SHADD16

MerryMage:
Verification: https://ideone.com/5q0yxm

Non-SSE. 

Algorithm derivation:

Since `x+y == ((x&y) << 1) + (x^y)`, to calculate `(x+y)>>1`, calculate `(x&y) + ((x^y) >> 1)`. Note that the latter algorithm allows for the lowermost bit to be easily masked away and discarded.

    (x&y) == aaaaaaaabbbbbbbbccccccccdddddddd
    ((x^y)>>1) & 0x7F7F7F7F == 0eeeeeee0fffffff0ggggggg0hhhhhhh

Note that simply adding the above two numbers is the correct behaviour for *unsigned* addition, but we need to sign extend the latter value in this *signed* implementation. We can calculate the required sign bit like so:

    (x^y) & 0x80808080 == e0000000f0000000g0000000h0000000

The sign bit can be xored into the result as xor is carryless addition.

Ping @Subv.


MerryMage:
Passes tests.

PR #51, opened 2016-11-27
VFP: Fixed the VCVT behavior when converting from unsigned 32-bit values

Subv:
Use a 64-bit register to hold the values so that we don't end up interpreting them as signed values.


MerryMage:
Two issues:

* Result needs to be clamped between 0 and 2^32-1.
* `FPExc_InvalidOp` needs to be raised when clamping happens. (i.e.: `FPSCR.IOC = 1`)

(See also FPToFixed() function in ARM manual.)

MerryMage:
LGTM

PR #53, opened 2016-11-30
microinstruction: Make use_count private

lioncash:
Makes the operation a part of the direct interface.

I've kept the name as is, but would it be preferable to rename it to `DecrementUseCount()`?


MerryMage:
The name is kind of intentionally misfitting. `DecrementRemainingUses` is only intended to be used by backends, while the other two are mainly for the frontend. This is because `use_count` has slightly different semantics in both cases.

In the frontend, it refers to the number of microinstructions that point to this. In the backend, it refers to the number of microinstructions that point to this that haven't been processed yet.

PR #60, opened 2016-12-11
Support stack unwinding on Windows

MerryMage:
This provides support for stack unwinding only on Windows, since that was the simplest interface to implement.

**Note for people from the future:**

*Supporting other debuggers:*

* GDB has a JIT interface: https://sourceware.org/gdb/onlinedocs/gdb/JIT-Interface.html (Annoyingly, one must implement an ELF and a DWARF emitter.)
* libunwind has a dynamic code registration interface: http://www.nongnu.org/libunwind/man/_U_dyn_register(3).html

*Supporting exceptions on non-Windows platforms*

* libsupcxx: Requires you to construct .eh_frame entries and to pass it to __register_frame.


src/common/assert.h
```diff
@@ -22,19 +22,20 @@ __declspec(noinline, noreturn)
 #endif
 static void assert_noinline_call(const Fn& fn) {
     fn();
+    throw "";
     exit(EXIT_FAILURE); // Keeps GCC's mouth shut about this actually returning
```
yuriks: I think if you have the throw there this exit shouldn't be necessary anymore.

src/backend_x64/unwind_windows.cpp
```diff
@@ -0,0 +1,202 @@
...
+    save_xmm128(29, 7, 0x30);
+    save_xmm128(35, 8, 0x40);
+    save_xmm128(41, 9, 0x50);
+    save_xmm128(47, 10, 0x60);
+    save_xmm128(54, 11, 0x70);
+    save_xmm128(63, 12, 0x80);
+    save_xmm128(72, 13, 0x90);
+    save_xmm128(81, 14, 0xA0);
+    save_xmm128(90, 15, 0xB0);
+    ret.prolog_size = 90;
```
yuriks: You have a counting error here which is breaking variable value display in frames below the JIT (not sure how it didn't break the actual stack trace too):

```diff
diff --git a/src/backend_x64/unwind_windows.cpp b/src/backend_x64/unwind_windows.cpp
index 8fa5276..d211d6b 100644
--- a/src/backend_x64/unwind_windows.cpp
+++ b/src/backend_x64/unwind_windows.cpp
@@ -137,12 +137,12 @@ static PrologueInformation GetPrologueInformation() {
     save_xmm128(35, 8, 0x40);
     save_xmm128(41, 9, 0x50);
     save_xmm128(47, 10, 0x60);
-    save_xmm128(54, 11, 0x70);
-    save_xmm128(63, 12, 0x80);
-    save_xmm128(72, 13, 0x90);
-    save_xmm128(81, 14, 0xA0);
-    save_xmm128(90, 15, 0xB0);
-    ret.prolog_size = 90;
+    save_xmm128(53, 11, 0x70);
+    save_xmm128(62, 12, 0x80);
+    save_xmm128(71, 13, 0x90);
+    save_xmm128(80, 14, 0xA0);
+    save_xmm128(89, 15, 0xB0);
+    ret.prolog_size = 89;

     ret.number_of_unwind_code_entries = ret.unwind_code.size();

```

yuriks: I think it'd be better to have the code keep track of the offset and change the first parameter to be the size of the instruction instead. Should make it easier to spot these issues, I think?

(Could also add the code in question as a comments to the end of each line like you had before.

```
+000  53                          push    rbx
+001  56                          push    rsi
+002  57                          push    rdi
+003  55                          push    rbp
+004  41 54                       push    r12
+006  41 55                       push    r13
+008  41 56                       push    r14
+00A  41 57                       push    r15
+00C  48 81 EC C8 00 00 00        sub     rsp,0C8h
+013  0F 29 74 24 20              movaps  xmmword ptr [rsp+20h],xmm6
+018  0F 29 7C 24 30              movaps  xmmword ptr [rsp+30h],xmm7
+01D  44 0F 29 44 24 40           movaps  xmmword ptr [rsp+40h],xmm8
+023  44 0F 29 4C 24 50           movaps  xmmword ptr [rsp+50h],xmm9
+029  44 0F 29 54 24 60           movaps  xmmword ptr [rsp+60h],xmm10
+02F  44 0F 29 5C 24 70           movaps  xmmword ptr [rsp+70h],xmm11
+035  44 0F 29 A4 24 80 00 00 00  movaps  xmmword ptr [rsp+80h],xmm12
+03E  44 0F 29 AC 24 90 00 00 00  movaps  xmmword ptr [rsp+90h],xmm13
+047  44 0F 29 B4 24 A0 00 00 00  movaps  xmmword ptr [rsp+0A0h],xmm14
+050  44 0F 29 BC 24 B0 00 00 00  movaps  xmmword ptr [rsp+0B0h],xmm15
```

yuriks: Also reversing the order of the entries fixes the rest of the variables, just for the record.

PR #61, opened 2016-12-12
Added cpuid detection library that's safe (no inline asm)

FernandoS27:
A very big plus the JIT compiler could use, is the ability to detect Instruction sets on runtime and make use of the best instructions for jit compiling like AVX and BMI. This, however, will require the JIT compiler to be refactored into the builder pattern to create a jit compiler that uses the best instructions available and avoids code bloating (adding conditionals everywhere).

So why use this code snippet instead of cpu_id in citra/dragonfly?

- flexibility: easier to mantain,can be extended for future processors.
- no inline-asm. Citra's cpu_id snippet relies on inline_asm and this is very dangerous since even if the instruction set is supported, the OS may have it disabled. Instead we use compiler intrinsics to let the compiler decide how the instruction should be used.
- encapsulation.
- supports MSVC, GCC and Clang.


MerryMage:
Thank you for your contribution.

We already have a well-maintained cpuid interface as part of Xbyak.

> no inline-asm. Citra's cpu_id snippet relies on inline_asm and this is very dangerous since even if the instruction set is supported, the OS may have it disabled. Instead we use compiler intrinsics to let the compiler decide how the instruction should be used.

I'm not sure you're aware but those two intrinsics are literally just `cpuid`.

FernandoS27:
yeah but intrinsics are managed by the compiler. I prefer intrinsics, inline_asm is always a practice that should be avoided. Thanks though.

are there any unassigned tasks?

MerryMage:
@FernandoS27 Perhaps you could visit the [#citra IRC](https://webchat.freenode.net/?channels=citra) and discuss things in real time.

PR #64, opened 2016-12-17
Implement USAD8 and USADA8, and fixes alloc not renamed on unwinding_windows.cpp

FernandoS27:
Implemented Sum of absolute differences with test cases, feel free to review


wwylele:
Is it because of the change of line break that the diff is so huge?

FernandoS27:
wtf happened? The diffs are very different to my last push, thankfully the other pull contains this changes as well.

PR #76, opened 2016-12-23
[RFC] coprocessor_util: Expose coprocessor decoding functions for library users

MerryMage:
Not too happy with this.

Would probably be useful for library users; though need it even be done? i.e.: Someone who write a coprocessor could probably do decoding themselves.


MerryMage:
Probably completely unnecessary if parsing is done on the library user's behalf.

MerryMage:
This is fine.

PR #80, opened 2016-12-31
Coprocessor Interface

MerryMage:
**What is this:**

A coprocessor interface for users of this library to define coprocessors. This includes cp15.

**Issues that arose:**

1. `std::array<u8, 8>` is currently used to shuttle around coprocessor instruction information. This is not as type-safe as I want it to be. If there's an alternative type-safe way of cramming that information into 64-bits, that would be preferable. Even though I don't like the current design very much, I can't think something better that doesn't unnecessarily enlarge the size of `IR::Inst` or `IR::Value`.
2. `EmitCoprocessorException` isn't implemented. This is intentional, it will be implemented when `UndefinedInstruction` and related callbacks are implemented in the future. It is currently handled the same way the UDF instruction is handled at the moment: with an `ASSERT`.

I'll let this sit here for a week or two until I'm fully happy with it.


MerryMage:
Probably will merge as-is with minor corrections.

src/frontend/ir/opcodes.inc
```diff
@@ -160,3 +160,12 @@ OPCODE(ExclusiveWriteMemory8,   T::U32,         T::U32,         T::U8
...
 OPCODE(ExclusiveWriteMemory32,  T::U32,         T::U32,         T::U32                          )
 OPCODE(ExclusiveWriteMemory64,  T::U32,         T::U32,         T::U32,         T::U32          )
+
+// Coprocessor
+OPCODE(CoprocInternalOperation, T::Void,        T::CoprocInfo                                   )
+OPCODE(CoprocSendOneWord,       T::Void,        T::CoprocInfo,  T::U32                          )
+OPCODE(CoprocSendTwoWords,      T::Void,        T::CoprocInfo,  T::U32,         T::U32          )
+OPCODE(CoprocGetOneWord,        T::U32,         T::CoprocInfo                                   )
+OPCODE(CoprocGetTwoWords,       T::U64,         T::CoprocInfo                                   )
+OPCODE(CoprocLoadWords,         T::U32,         T::CoprocInfo,  T::U32                          )
```
MerryMage: `CoprocLoadWords` and `CoprocStoreWords` do not return anything.

FernandoS27:
LGTM.

PR #91, opened 2017-02-15
Ranged cache invalidation

lynn:
This implements `EmitX64::InvalidateCacheRange`, which invalidates the cache range passed to it. `Jit::Impl` tracks a queue of ranges to invalidate, instead of a single boolean `clear_cache_required`.

Also, I renamed `code_ptr` in `BlockDescriptor` to `entrypoint` ~~to make this PR harder to merge~~ for readability.

### DIFF
--- a/src/backend_x64/interface_x64.cpp
+++ b/src/backend_x64/interface_x64.cpp
@@ -5,6 +5,7 @@
  */
 
 #include <memory>
+#include <queue>
 
 #include <fmt/format.h>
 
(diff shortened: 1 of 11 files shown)


src/backend_x64/interface_x64.cpp
```diff
@@ -35,30 +36,32 @@ struct Jit::Impl {
             , jit_state()
             , emitter(&block_of_code, callbacks, jit)
             , callbacks(callbacks)
+            , invalid_cache_ranges()
```
lioncash: This doesn't need to be present, as `std::queue`'s constructor executes regardless.

PR #92, opened 2017-02-16
[Translator] Microoptimize computing info only when needed.

inquisitio:
This is similar to what other instructions do, e.g. ADC.


MerryMage:
Thanks but this doesn't match the style of most of the rest of the codebase. Generally operations related to  decoding occur outside of the `if` statement and operations relating to execution occur within the `if` statement.

Generally, the translators should match ARM documentation as far as is reasonable.

PR #96, opened 2017-02-23
CMakeLists: Change set into a list append

lioncash:
Same thing, less repetition.

Also puts quotes around the added path to correctly handle any potential spaces in paths.


CMakeLists.txt
```diff
@@ -29,8 +29,8 @@ if ("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}")
     message(SEND_ERROR "In-source builds are not allowed.")
 endif()
 
-# Signify where the module directories are
```
yuriks: https://www.youtube.com/watch?v=XVHkO_uT95E

PR #100, opened 2017-03-04
CMakeLists: Make boost an interface library target

lioncash:
Gets rid of the use of a non-target include and makes libraries explicitly link against the identifier name in order to get includes.


CMakeLists.txt
```diff
@@ -109,7 +109,8 @@ else()
         message(FATAL_ERROR "Please provide a path to a boost installation using Boost_INCLUDE_DIRS")
     endif()
 endif()
-include_directories(${Boost_INCLUDE_DIRS})
+add_library(boost INTERFACE IMPORTED)
+set_property(TARGET boost PROPERTY INTERFACE_INCLUDE_DIRECTORIES ${Boost_INCLUDE_DIRS})
```
yuriks: Does
```
target_include_directories(boost INTERFACE ${Boost_INCLUDE_DIRS})
```
work instead?

lioncash: That isn't applicable for imported targets, unfortunately.

FernandoS27:
Makes sense especialy since we mainly use boost optional and boost variant.

PR #104, opened 2017-03-09
externals: Update catch to 1.8.2

lioncash:
Keeps the unit-testing library up to date. The previous version of catch being used was 1.5.6, which is quite out of date.

catch lists all its feature changes on its [release page](https://github.com/philsquared/Catch/releases) for those interested.

edit: fails on windows at the moment because of virtual destructor warnings, which are fixed in this [commit](https://github.com/philsquared/Catch/commit/d5a5883a10918d0f2c9aed5c8c18b33a7f271b0f). Just need to wait until the next release.


lioncash:
There we go, 1.8.2 fixed the failures due to warnings.

PR #106, opened 2017-05-17
Link against static fmtlib when DYNARMIC_NO_BUNDLED_FMT is set

jroweboy:
When including fmtlib as a header only library in dynarmic, downstream
projects cannot include fmtlib as a static library without getting
linker errors. This change make dynarmic use the static library version
of fmtlib when DYNARMIC_NO_BUNDLED_FMT is set

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -115,7 +115,7 @@ target_compile_options(dynarmic PRIVATE ${DYNARMIC_CXX_FLAGS})
 target_link_libraries(dynarmic
     PRIVATE
         boost
-        fmt-header-only
+        fmt
         xbyak
         $<$<BOOL:DYNARMIC_USE_LLVM>:${llvm_libs}>
 )


src/CMakeLists.txt
```diff
@@ -115,10 +115,15 @@ target_compile_options(dynarmic PRIVATE ${DYNARMIC_CXX_FLAGS})
 target_link_libraries(dynarmic
     PRIVATE
         boost
-        fmt-header-only
         xbyak
         $<$<BOOL:DYNARMIC_USE_LLVM>:${llvm_libs}>
 )
+# Link against static fmtlib if the downstream project provides fmtlib
+if (DYNARMIC_NO_BUNDLED_FMT)
+target_link_libraries(dynarmic PRIVATE fmt)
```
lioncash: indentation

MerryMage:
Would it be cleaner to just use `fmt`?

jroweboy:
Updated to use `fmt` instead of `fmt-header-only`

PR #112, opened 2017-10-18
Improved performance of reg_alloc

huwpascoe:
Noticed slight improvement in profiling. Reduced hotpath samples.
The linked list is probably better suited for these random queries.

std::vector:
https://godbolt.org/g/nU8Pg6

std::forward_list:
https://godbolt.org/g/r2o1FH


huwpascoe:
Did not see that dependency, lol. Oh visual studio....

PR #117, opened 2017-11-12
Common: Delete Pool's copy constructor and copy/move assignment operators

lioncash:
Copy constructors need the `const` qualifier on its parameter.

Also, should the copy-assignment and move-assignment operators be deleted as well?


MerryMage:
> Also, should the copy-assignment and move-assignment operators be deleted as well?

Absolutely. Sorry, my mistake.

lioncash:
Done~

PR #123, opened 2017-12-12
Sync

CDAGaming:


FernandoS27:
this looks like a nice case, I never expected arguments to be held on xmm registers instead, probably when redesigning the IR language a code generation tree could be used on the target and the right path could be picked on the context of the other microcodes of the ir (either all generated in GPR or all in SSE based in block context).

Sadly I wouldn't suggest this to be merged since it seems to be an small optimization patch. Thanks for trying though.

CDAGaming:
@FernandoS27 This PR was a mistake (Was trying to target my branch instead of this)
However, this was already merged back into master by merry in https://github.com/MerryMage/dynarmic/commit/13fa10e743284a2f1dec0139dabd9f20bc5b8ac7

PR #124, opened 2017-12-18
emit_x64: InvalidateCacheRanges: Erase blocks after iterating through the map

Phanto-m:
`icl::interval_map` doesn't preserve borders of inserted intervals so when invalidating a region where blocks overlap you would only erase the overlap in `block_ranges`
Storing `LocationDescriptor`s and erasing them after iterating makes it possible to erase the correct range at once
Also, `block_ranges.clear();` in `EmitX64::ClearCache()`


MerryMage:
Nice catch.

PR #133, opened 2018-01-22
A64: Implement CLZ

lioncash:


src/backend_x64/emit_x64.cpp
```diff
@@ -1327,6 +1327,31 @@ void EmitX64<JST>::EmitCountLeadingZeros(EmitContext& ctx, IR::Inst* inst) {
...
+       code->lzcnt(result, source);
+
+       ctx.reg_alloc.DefineValue(inst, result);
+   } else {
+       Xbyak::Reg64 source = ctx.reg_alloc.UseScratchGpr(args[0]).cvt64();
+       Xbyak::Reg64 result = ctx.reg_alloc.ScratchGpr().cvt64();
+
+       // The result of a bsr of zero is undefined, but zf is set after it.
+       code->bsr(result, source);
+       code->mov(source, 0xFFFFFFFFFFFFFFFF);
```
MerryMage: Smaller codesize:

    code->mov(source.cvt32(), 0xFFFFFFFF);
    code->cmovz(result.cvt32(), source.cvt32());
    code->neg(result.cvt32());
    code->add(result.cvt32(), 63);

Upper 32 bits of a register are always zeroed after a 32 bit operation on x64.

src/frontend/ir/ir_emitter.h
```diff
@@ -135,7 +135,9 @@ class IREmitter {
     U32 ByteReverseWord(const U32& a);
     U16 ByteReverseHalf(const U16& a);
     U64 ByteReverseDual(const U64& a);
-    U32 CountLeadingZeros(const U32& a);
+    U32 CountLeadingZeros32(const U32& a);
+    U64 CountLeadingZeros64(const U64& a);
```
MerryMage: Do you prefer the explicitness of `CountLeadingZeros{32,64}` versus calling them all `CountLeadingZeros` and letting overload resolution sort it out?

lioncash: ah, that was a carry-over the emitter function naming, I'll correct that.

src/backend_x64/emit_x64.cpp
lioncash: Ah right, forgot about that. Good catch.

src/frontend/ir/ir_emitter.h
lioncash: ah, that was a carry-over the emitter function naming, I'll correct that.

src/backend_x64/emit_x64.cpp
lioncash: Forgot about that. Good catch.

lioncash:
@MerryMage should be all addressed now

PR #135, opened 2018-01-24
A64: Implement RBIT

lioncash:


src/frontend/A64/translate/impl/data_processing_register.cpp
```diff
@@ -29,6 +29,47 @@ bool TranslatorVisitor::CLS_int(bool sf, Reg Rn, Reg Rd) {
     return true;
 }
 
+bool TranslatorVisitor::RBIT_int(bool sf, Dynarmic::A64::Reg Rn, Dynarmic::A64::Reg Rd) {
```
MerryMage: `Dynarmic::A64::` unnecessary

lioncash: Ah, Xcode autocomplete owned me here.

PR #140, opened 2018-01-25
Implemented UMULH and SMULH instructions

FernandoS27:


src/frontend/A64/translate/impl/data_processing_multiply.cpp
```diff
@@ -79,6 +79,17 @@ bool TranslatorVisitor::UMSUBL(Reg Rm, Reg Ra, Reg Rn, Reg Rd) {
     return true;
 }
 
+bool TranslatorVisitor::UMULH(Reg Rm, Reg Rn, Reg Rd) {
+
```
lioncash: This newline should be removed

src/backend_x64/emit_x64_data_processing.cpp
```diff
@@ -1255,5 +1255,16 @@ void EmitX64::EmitCountLeadingZeros64(EmitContext& ctx, IR::Inst* inst) {
    }
 }
 
+void EmitX64::EmitUnsignedMultiplyHigh64(EmitContext& ctx, IR::Inst* inst) {
+   auto args = ctx.reg_alloc.GetArgumentInfo(inst);
+
+   ctx.reg_alloc.ScratchGpr({HostLoc::RDX});
+   ctx.reg_alloc.UseScratch(args[0], HostLoc::RAX);
+   OpArg op_arg = ctx.reg_alloc.UseOpArg(args[1]);
+   code->mul(*op_arg);
```
lioncash: I'd imagine the signed equivalent would likely just use a mov and an imul, i.e. something like

```
mov rax, rsi
imul rdi
mov rax, rdx
```

if you feel like implementing SMULH as well.

FernandoS27: didn't figure you could do that as well with imul xD, done.

lioncash:
This needs a rebase

src/frontend/ir/opcodes.inc
```diff
@@ -122,6 +122,8 @@ OPCODE(ByteReverseHalf,         T::U16,         T::U16
 OPCODE(ByteReverseDual,         T::U64,         T::U64                                          )
 OPCODE(CountLeadingZeros32,     T::U32,         T::U32                                          )
 OPCODE(CountLeadingZeros64,     T::U64,         T::U64                                          )
+OPCODE(UnsignedMultiplyHigh64,  T::U64,         T::U64                                          )
```
lioncash: The reason the assertion fails on the CI is because there's only one argument provided here, but two are being used in `EmitSignedMultiplyHigh64` and `EmitUnsignedMultiplyHigh64`

FernandoS27: I'll look into it

FernandoS27:
"Rebased" (actualy reimplemented in a new branch) and fixed the assertion issue.

PR #142, opened 2018-01-25
Implemented Vector Instructions: EOR, ORR, ORN

FernandoS27:


src/frontend/ir/opcodes.inc
```diff
@@ -175,6 +175,9 @@ OPCODE(VectorAdd16,             T::U128,        T::U128,        T::U128
 OPCODE(VectorAdd32,             T::U128,        T::U128,        T::U128                         )
 OPCODE(VectorAdd64,             T::U128,        T::U128,        T::U128                         )
 OPCODE(VectorAnd,               T::U128,        T::U128,        T::U128                         )
+OPCODE(VectorOr,                T::U128,        T::U128,        T::U128                         )
+OPCODE(VectorExclusiveOr,       T::U128,        T::U128,        T::U128                         )
```
MerryMage: We've been using Eor for exclusive or.

src/backend_x64/emit_x64_vector.cpp
```diff
@@ -121,6 +121,26 @@ void EmitX64::EmitVectorAnd(EmitContext& ctx, IR::Inst* inst) {
...
+    EmitVectorOperation(code, ctx, inst, &Xbyak::CodeGenerator::pxor);
+}
+
+void EmitX64::EmitVectorNot(EmitContext& ctx, IR::Inst* inst) {
+    auto args = ctx.reg_alloc.GetArgumentInfo(inst);
+
+    Xbyak::Xmm xmm_a = ctx.reg_alloc.UseScratchXmm(args[0]);
+    Xbyak::Xmm xmm_b = ctx.reg_alloc.ScratchXmm();
+
+    code->ptest(xmm_b, xmm_b);
```
MerryMage: `ptest` is SSE4.1 only. It's also not the most efficient way to set a register to all ones (the CPU), costing two execution ports. On Haswell, it has a 2 cycle latency

The most optimal way to set a register to all ones is `pcmpeqw`, which costs zero execution ports when used for this purpose and has a zero cycle latency.

src/frontend/A64/translate/impl/simd_three_same.cpp
```diff
@@ -74,5 +74,44 @@ bool TranslatorVisitor::AND_asimd(bool Q, Vec Vm, Vec Vn, Vec Vd) {
...
+}
+
+bool TranslatorVisitor::ORN_asimd(bool Q, Vec Vm, Vec Vn, Vec Vd) {
+    const size_t datasize = Q ? 128 : 64;
+
+    auto operand1 = V(datasize, Vn);
+    auto operand2 = V(datasize, Vm);
+
+    auto result = ir.VectorOr(operand1, ir.VectorNot(operand2));
+
```
MerryMage: Insert this comment:

`// TODO: This does not zero the upper 64 bits when datasize == 64. This may break future optimization passes.`

src/backend_x64/emit_x64_vector.cpp
FernandoS27: I wasn't sure which instruction we were using before, got it.

FernandoS27:
Updated.

lioncash:
Can you put the changes for each of those in their respective original commit instead of creating a new commit for them all? It'll make the commit history nicer.

MerryMage:
That comment applied to only ORN_asimd, the rest are fine.

FernandoS27:
me doesn't know how to amend specific commits xD, so icreated a new one.

FernandoS27:
done, I just squashed everything to a single commit.

PR #145, opened 2018-01-26
A64: Implement EOR3 and BCAX

lioncash:
I expect BCAX to fail in tests for some obscure reason. This is just to see if the buildbot also chokes on it.


FernandoS27:
hmn not sure why BCAX should fail. It seems the appveyor run liked it.

PR #152, opened 2018-01-27
Implemented BSL, BIC, BIT and BIF vector instructions

FernandoS27:


MerryMage:
Please use the naming convention in the manual, because reading the code is slightly confusing with the use of different names.

FernandoS27:
Fixed, the code now matches the documentation.

MerryMage:
> Should we be doing this for every instruction?

Only the instructions that need it (if the upper 64 bits are not guaranteed to be zero).

PR #155, opened 2018-01-28
A64: Implement CRC32C

lioncash:
The regular CRC32 variant actually isn't implemented [properly](https://github.com/unicorn-engine/unicorn/blob/master/qemu/target-arm/helper-a64.c#L427) in unicorn, because unicorn is very good software. So once I fix that I'll be PRing the regular CRC32 variant.


FernandoS27:
So unicorn is not fault free, *sight*

PR #159, opened 2018-02-01
A64: Implement AESIMC and AESMC

lioncash:


src/backend_x64/emit_x64_aes.cpp
```diff
@@ -0,0 +1,65 @@
...
+namespace Dynarmic::BackendX64 {
+
+using namespace Xbyak::util;
+    
+using MixColumnsFn = void(Common::AESState&, const Common::AESState&);
+
+static void EmitMixColumns(std::array<Argument, 3> args, EmitContext& ctx, BlockOfCode& code,
+                           IR::Inst* inst, MixColumnsFn fn) {
+    Xbyak::Xmm input = ctx.reg_alloc.UseXmm(args[0]);
+    Xbyak::Xmm output = ctx.reg_alloc.ScratchXmm();
```
MerryMage: It's not necessary to allocate a register for this.

src/backend_x64/emit_x64_aes.cpp
```diff
@@ -0,0 +1,65 @@
...
+    ctx.reg_alloc.EndOfAllocScope();
+
+    ctx.reg_alloc.HostCall(nullptr);
+    code.sub(rsp, 16 * 2 + ABI_SHADOW_SPACE);
+    code.lea(code.ABI_PARAM1, ptr[rsp + ABI_SHADOW_SPACE]);
+    code.lea(code.ABI_PARAM2, ptr[rsp + ABI_SHADOW_SPACE + 16]);
+
+    code.movaps(xword[code.ABI_PARAM1], output);
```
MerryMage: You can leave that memory uninitialised.

src/backend_x64/emit_x64_aes.cpp
```diff
@@ -0,0 +1,64 @@
...
+    // Free memory
+    code.add(rsp, stack_space + ABI_SHADOW_SPACE);
+
+    ctx.reg_alloc.DefineValue(inst, xmm0);
+}
+
+void EmitX64::EmitAESInverseMixColumns(EmitContext& ctx, IR::Inst* inst) {
+     auto args = ctx.reg_alloc.GetArgumentInfo(inst);
+
+    if (code->DoesCpuSupport(Xbyak::util::Cpu::tAVX)) {
```
MerryMage: Sorry I didn't spot this earlier, this should be `tAESNI` (`aesimc` is an AES-NI instruction. `vaesimc`, on the other hand, requires both AES-NI and AVX.)

lioncash: Good catch. Fixed.

PR #165, opened 2018-02-06
A64: Implement ZIP1

lioncash:


MerryMage:
Just one question: Why is `Upper` in the name? It looks like an interleave of the lower half of the vector is performed.

PR #173, opened 2018-03-21
bit_util: Do nothing in RotateRight if the rotation amount is zero

lioncash:
Without this sanitizing it's possible to perform a shift with a shift amount that's the same size as the type being shifted. This actually occurs when decoding ORR variants.

We could get fancier here and make this branchless, but we don't really use RotateRight in any performance intensive areas.


wwylele:
I think I have read some discussion saying that gcc/clang (?) can detect this sanitizing and still compile it to a single rotation instruction.

MerryMage:
Comparison should probably come after the `amount %= BitSize<T>();` line.

lioncash:
@wwylele Yep GCC compiles it down to an AND and a ROR. I was speaking more in general than specific compilers about being fancy though.

PR #177, opened 2018-03-28
A64: system: Use an enum class for MRS/MSR register encodings

lioncash:
Reduces the need to manually write out the register bit encodings repeatedly.


lioncash:
No functionality changed in the amendment. The only difference is the enum elements are alphabetical now

PR #186, opened 2018-04-02
A64: Implement USUBW/USUBW2

lioncash:


lioncash:
Note: The failure on this one isn't related to USUBW/USUBW2

PR #190, opened 2018-04-03
A64: Implement RBIT (vector)

lioncash:
x86 is just straight-up goofy when it comes to bit reversal capabilities


src/backend_x64/emit_x64_vector.cpp
```diff
@@ -1276,6 +1276,49 @@ void EmitX64::EmitVectorPopulationCount(EmitContext& ctx, IR::Inst* inst) {
     });
 }
 
+void EmitX64::EmitVectorReverseBits(EmitContext& ctx, IR::Inst* inst) {
+    auto args = ctx.reg_alloc.GetArgumentInfo(inst);
+
+    const Xbyak::Xmm data = ctx.reg_alloc.UseScratchXmm(args[0]);
+    const Xbyak::Xmm high_nibble_reg = ctx.reg_alloc.ScratchXmm();
```
lioncash: This should likely have a better name, though I wasn't sure what to call it. It initially starts out being a high nibble mask register, but is then used for other purposes, depending on the SSE feature branch taken.

src/backend_x64/emit_x64_vector.cpp
```diff
@@ -1276,6 +1276,49 @@ void EmitX64::EmitVectorPopulationCount(EmitContext& ctx, IR::Inst* inst) {
...
+    code.movdqa(high_nibble_reg, code.MConst(xword, 0xF0F0F0F0F0F0F0F0, 0xF0F0F0F0F0F0F0F0));
+    code.pand(high_nibble_reg, data);
+    code.pxor(data, high_nibble_reg);
+    code.psrld(high_nibble_reg, 4);
+
+    if (code.DoesCpuSupport(Xbyak::util::Cpu::tSSSE3)) {
+        // High lookup
+        const Xbyak::Xmm high_reversed_reg = ctx.reg_alloc.ScratchXmm();
+        code.movdqa(high_reversed_reg, code.MConst(xword, 0xE060A020C0408000, 0xF070B030D0509010));
+        code.pshufb(high_reversed_reg, data);
```
MerryMage: `code.pshufb(high_reversed_reg, code.MConst(xword, 0xE060A020C0408000, 0xF070B030D0509010));`
saves having to allocate temporary registers

PR #194, opened 2018-04-04
A64: Implement URSHR (vector) and URSRA (vector)

lioncash:


lioncash:
(sync just moved the initial brace placement on the enum class; no behavior was changed)

PR #204, opened 2018-04-10
A64: Implement SM3PARTW2

lioncash:
After this, I'll likely add a similar `RotateLeft` non-vector helper to make code more self documenting (instead of inverting the values accordingly to work in terms of rotate right). The only reason I've excluded that is I'd like to also make `RotateRight` accept a plain `u8` instead of an `IR::U8`, which would clean up rotation call sites a bit. I felt including both that cleanup and the implementation of an instruction in the same PR was a little cluttery.


MerryMage:
> I'll likely add a similar RotateLeft non-vector helper to make code more self documenting (instead of inverting the values accordingly to work in terms of rotate right). 

sgtm 

> I'd like to also make RotateRight accept a plain u8 instead of an IR::U8, which would clean up rotation call sites a bit. 

I’m also fine with this

lioncash:
Nice 👌

PR #217, opened 2018-04-16
A64: Implement CMGE (zero)'s vector variant

lioncash:
Should this and CMGT be zeroing the upper half of the vector when the data size is 64, or is this something only CMEQ needs to do? If those two should be doing it as well, I can amend this PR to fix that.


MerryMage:
Yes, please zero.

lioncash:
Done

PR #230, opened 2018-04-20
emit_x64_vector: Generify variable shift functions

lioncash:
Gets rid of a tiny bit of duplicated code.


src/backend_x64/emit_x64_vector.cpp
```diff
@@ -790,63 +790,43 @@ void EmitX64::EmitVectorLogicalShiftRight64(EmitContext& ctx, IR::Inst* inst) {
     ctx.reg_alloc.DefineValue(inst, result);
 }
 
+template <typename T = void>
```
MerryMage: Is the default necessary?

lioncash: Oh, I have that removed locally, guess I forgot to amend my commit. Will correct that

PR #244, opened 2018-04-27
A64: Implement CMTST's scalar variant

lioncash:
This can just be done in terms of the vector equivalent for now


lioncash:
Failure isn't related to this implementation

PR #245, opened 2018-04-27
emit_x64_vector: Vectorize fallback path in EmitVectorMaxU16()

lioncash:


lioncash:
Failure isn't related to this vectorization

PR #256, opened 2018-05-03
emit_x64_vector_floating_point: Deduplicate accurate NaN handling code

lioncash:
Allows the code to both be used from the 32 bit and 64 bit operations without duplicating code. If there's a more preferable name than `HandleNaNs`, I change it to that instead.


MerryMage:
Perhaps also move the `ptest` code into the function as well (just that `if, else` block).

lioncash:
Done

PR #261, opened 2018-05-04
emit_x64_vector: Use VPOPCNTB in EmitVectorPopulationCount() if AVX-512 BITALG is available

lioncash:


lioncash:
(failure is related to fdiv)

PR #276, opened 2018-05-14
emit_x64_vector: Emit VPMULLQ in EmitVectorMultiply64 on AVX-512{DQ, VL} capable CPUs

lioncash:
Shortens code-gen down to a single instruction in the 64-bit path.


lioncash:
(failure related to fdiv)

PR #279, opened 2018-05-16
fuzz_with_unicorn: Remove exclusion of FMOV (imm) for FP-16 floats

lioncash:
Qemu, or rather, Unicorn now supports FP-16, since I backported support for the recent changes to mainline Qemu relating to FP-16 support.


lioncash:
(failure related to FDIV)

PR #293, opened 2018-07-07
A64: Implement FACGE and FACGT's vector single/double precision variants

lioncash:


lioncash:
Failure is related to FDIV

PR #295, opened 2018-07-07
A64: Implement FMAX and FMIN's vector single and double precision variants.

lioncash:


lioncash:
Right, forgot that Aarch64 handles zeros in min{pd|ps}/max{pd|ps} a little differently.

MerryMage:
(A64 also handles NaNs differently iirc.)

MerryMage:
Pushed with zeros fix as [6c37c31...4c3ca51](https://github.com/MerryMage/dynarmic/compare/6c37c311de08c073a767778c1d125e5a352bbe3c...4c3ca51e8668162e27f0b852215c597793133801).

PR #296, opened 2018-07-09
simd_scalar_two_register_misc: Handle 64-bit case in SCVTF and UCVTF's scalar double/single-precision variant

lioncash:
Avoids falling back to the interpreter in these two cases.


lioncash:
Failure related to FDIV

PR #298, opened 2018-07-10
floating_point_conversion_integer: Handle remaining fallback cases in SCVTF_float_int and UCVTF_float_int

lioncash:
Handles the I64 → F32 cases for both signednesses, which is the only fallback case left, since the changes in #297


lioncash:
Failure seems unrelated

PR #299, opened 2018-07-12
A64: Implement UCVTF (vector, integer)'s double/single-precision variant

lioncash:


MerryMage:
Requires correction as in dbfdb2506a8e38a75888bef1bec46964ff91af65

lioncash:
Done

PR #301, opened 2018-07-13
inst_gen: Compress loop into std::any_of in IsInvalidInstruction()

lioncash:


lioncash:
Failure seems unrelated

PR #302, opened 2018-07-13
tests: Use type aliases to centralize type definitions

lioncash:
Centralizes all register and vector array definitions to a single set of aliases, so if these are ever changed (for whatever reason), then the rest of the testing code will follow suit without the need to manually change them. That, and it avoids needing to type out the array type and sizes repeatedly, and just lets one write a regular encompassing type name.


lioncash:
Failure seems unrelated

PR #303, opened 2018-07-13
A64: Implement scalar double/single-precision variants of FACGE, FACGT, FCMEQ, FCMGE, FCMGT

lioncash:


lioncash:
Failure seems to be unrelated as well

PR #311, opened 2018-07-16
 emit_x64_vector{_floating_point}: Add helper alias for sizing arrays relative to vector width

lioncash:
Avoids needing to remember to specify the proper size of the arrays, all that's needed is to specify the type of the array and the size will automatically be deduced from it. This helps prevent potential oversized or undersized arrays from being specified.


MerryMage:
This would also be useful in `emit_x64_vector_floating_point.cpp`.

lioncash:
Amended. Made the name a little more general for use within both places and moved the type alias to `emit_x64.h`

PR #317, opened 2018-07-17
emit_x64: Remove FPSCR_RoundTowardsZero() virtual function from EmitContext struct

lioncash:
This code was bugged in that we were comparing if the rounding mode was not equal to rounding towards zero. Fortunately, however, nothing uses this function anymore, and there's already the more general `FPSCR_RMode()` available, so this can be removed entirely.


lioncash:
failure is unrelated to the code itself (wholly CI-related)

PR #321, opened 2018-07-21
A64: The A64SetTPIDR IR instruction writes to a system register and should not be eliminated by the dead code elimination pass.

Subv:
Previously this instruction was always eliminated, resulting in incorrect values for TPIDR_EL0.

### DIFF
--- a/src/frontend/ir/microinstruction.cpp
+++ b/src/frontend/ir/microinstruction.cpp
@@ -152,6 +152,15 @@ bool Inst::WritesToCPSR() const {
     }
 }
 
+bool Inst::WritesToSystemRegister() const {
+    switch (op) {
+    case Opcode::A64SetTPIDR:
+        return true;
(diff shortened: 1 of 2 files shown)


src/frontend/ir/microinstruction.cpp
```diff
@@ -145,6 +145,7 @@ bool Inst::WritesToCPSR() const {
     case Opcode::A32SetGEFlags:
     case Opcode::A32SetGEFlagsCompressed:
     case Opcode::A64SetNZCV:
+    case Opcode::A64SetTPIDR:
```
lioncash: This should likely go in its own function named something like `WritesToSystemRegister()` (which gives a home for other system-related registers in the future), which would be added to the checks in `MayHaveSideEffects()` instead, since the TPIDR doesn't have anything to do with the CPSR.

PR #329, opened 2018-07-22
FPRSqrtEstimate: Deduplicate array bounds

lioncash:
Dehardcodes a few constants in the loops.


src/common/fp/op/FPRSqrtEstimate.cpp
```diff
@@ -22,16 +22,18 @@ namespace Dynarmic::FP {
...
+        LUT result{};
+        for (u64 i = 128; i < result.size(); i++) {
             u64 a = i;
             if (a < 256) {
                 a = a * 2 + 1;
             } else {
                 a = (a | 1) * 2;
             }
-            u64 b = 512;
+            u64 b = result.size();
```
MerryMage: `512` here is not related to the size of the array.

lioncash: amended

PR #330, opened 2018-07-22
A64: Implement FMLA, FMLS, and FMUL (by element)'s scalar double/single-precision variant

lioncash:


MerryMage:
Will merge when I write a proper fallback for FPMulAdd.

PR #342, opened 2018-07-29
A64: Implement SMAXP, SMINP, UMAXP, and UMINP

lioncash:
For the time being, we can just do a naive implementation which avoids falling back to the interpreter a bit. Horizontal operations aren't necessarily x86 SIMD's forte anyways.


MerryMage:
Fine with me.

PR #347, opened 2018-08-13
A64: Implement UQADD/UQSUB and lingering vector variants of SQADD/SQSUB

lioncash:


src/backend_x64/emit_x64_saturation.cpp
```diff
@@ -77,6 +77,45 @@ void EmitSignedSaturatedOp(BlockOfCode& code, EmitContext& ctx, IR::Inst* inst)
...
+    code.mov(addend, int_max);
+
+    if constexpr (size < 64) {
+        code.cmovae(addend.cvt32(), op_result.cvt32());
+    } else {
+        code.cmovae(addend, op_result);
+    }
+
+    if (overflow_inst) {
+        Xbyak::Reg overflow = ctx.reg_alloc.ScratchGpr();
```
lioncash: The above `op_result` register can be reused here instead of getting another scratch register. Would it be preferable to do that, or leave it as is for readability?

lioncash:
Failure seems unrelated

src/backend_x64/emit_x64_saturation.cpp
MerryMage: Leaving it as-is for readability is fine.

src/backend_x64/emit_x64_saturation.cpp
```diff
@@ -77,6 +77,45 @@ void EmitSignedSaturatedOp(BlockOfCode& code, EmitContext& ctx, IR::Inst* inst)
...
+        code.setc(overflow.cvt8());
```
MerryMage: `setb` perhaps? (yeah it's the exact same opcode but it matches `cmovae` above, which helps with not having to think too hard.)

lioncash: Done

lioncash:
failure also not related

PR #350, opened 2018-08-16
A64: Implement FMAXP, FMAXNMP, FMINP, and FMINNMP's vector variants

lioncash:
We can just implement these in terms of scalars for the time being.


src/frontend/A64/translate/impl/simd_three_same.cpp
```diff
@@ -247,6 +247,61 @@ bool PairedMinMaxOperation(TranslatorVisitor& v, bool Q, Imm<2> size, Vec Vm, Ve
...
+    const size_t datasize = Q ? 128 : 64;
+    const size_t elements = datasize / esize;
+    const size_t boundary = elements / 2;
+
+    const IR::U128 operand1 = v.V(datasize, Vn);
+    const IR::U128 operand2 = v.V(datasize, Vm);
+    IR::U128 result = v.ir.ZeroVector();
+
+    const auto fn = [&] {
+        switch (op) {
```
MerryMage: Might as well just pass the mfp in as an argument.

lioncash: Done

PR #352, opened 2018-08-16
preliminary changes to add aarch64 backend

BreadFish64:
I made several changes to the dolphin emitter so it compiles in dynarmic, mostly fixing warnings
diff here:
https://github.com/dolphin-emu/dolphin/compare/master...BreadFish64:dynarmic?expand=1


src/CMakeLists.txt
```diff
@@ -205,52 +205,59 @@ add_library(dynarmic
...
-         backend_x64/emit_x64_sm4.cpp
-         backend_x64/emit_x64_vector.cpp
-         backend_x64/emit_x64_vector_floating_point.cpp
-         backend_x64/hostloc.cpp
-         backend_x64/hostloc.h
-         backend_x64/jitstate_info.h
-         backend_x64/oparg.h
-         backend_x64/reg_alloc.cpp
-         backend_x64/reg_alloc.h
+         backend/X64/a32_emit_x64.cpp
```
MerryMage: Inconsistency in case of `x64`. This would break in case-sensitive file systems.

I would prefer lowercase, which would match the case in all the `#include`s.

src/backend/A64/dolphin_emitter/a64_emitter.h
```diff
@@ -0,0 +1,1144 @@
...
+
+#include <bitset>
+#include <cstring>
+#include <functional>
+
+#include "arm_common.h"
+#include "code_block.h"
+#include "common/assert.h"
+#include "common/common_types.h"
+//#include "Common.h"
```
MerryMage: No commented code please.

src/CMakeLists.txt
```diff
@@ -205,52 +205,59 @@ add_library(dynarmic
...
-        target_sources(dynarmic PRIVATE backend_x64/exception_handler_windows.cpp)
+        target_sources(dynarmic PRIVATE backend/X64/exception_handler_windows.cpp)
     else()
-        target_sources(dynarmic PRIVATE backend_x64/exception_handler_generic.cpp)
+        target_sources(dynarmic PRIVATE backend/X64/exception_handler_generic.cpp)
     endif()
+elseif(ARCHITECTURE_Aarch64)
+    target_sources(dynarmic PRIVATE
+         backend/A64/dolphin_emitter/a64_emitter.cpp
+		 backend/A64/dolphin_emitter/a64_emitter.h
```
MerryMage: Indentation.

src/common/bit_util.h
```diff
@@ -83,6 +83,22 @@ constexpr bool Bit(const T value) {
     return Bit<T>(bit_position, value);
 }
 
+/// Reinterpret objects of one type as another by bit-casting between object representations.
+template <typename To, typename From>
```
MerryMage: This already exists in `cast_util.h`

lioncash:
There's a massive amount of style discrepancies and tab/space mismatches

src/backend/A64/dolphin_emitter/arm_common.h
```diff
@@ -0,0 +1,26 @@
+// Copyright 2014 Dolphin Emulator Project / 2018 dynarmic project
+// Licensed under GPLv2+
+// Refer to the license.txt file included.
+
+#include "common/common_types.h"
+
+enum CCFlags {
```
MerryMage: Namespace this file. Try not to add anything to the global namespace; we’re a library.

src/CMakeLists.txt
lioncash: `x64` like the name already had.

lioncash:
The namespace should also not be `Dolphin`, it should be all unified under the `Dynarmic::BackendA64`, namespace, just like everything on the x64 platform is mostly unified under `Dynarmic::BackendX64`. I'd also like to state that I do not think adding another platform while the x64 one is still in flux is a good idea.

BreadFish64:
@lioncash I'm aware of the style issues, I asked Merry what to do on discord, I'll probably get to it tomorrow

MerryMage:
tbqh I don’t like Dolphin’s emitter; but I am willing to go with it initially then rip it out later.

If there is to be A64 backend work it probably should be on a branch and not on master. I would also prefer to have a build on CI (I have an example branch of that).

src/backend/A64/dolphin_emitter/a64_emitter.h
BreadFish64: oops

BreadFish64: oops

BreadFish64:
Ah I just noticed I was using my own master branch, did you mean that, or a branch on this repo?

BreadFish64:
I'm totally impartial when it comes to the emitter, another one that came up in discussion was [vixl](https://git.linaro.org/arm/vixl.git/about/) but this makes me wary:
```VIXL was developed for JavaScript engines so a number of features from A64 were deemed unnecessary:

Limited rounding mode support for floating point.

Limited support for synchronisation instructions.

Limited support for system instructions.

A few miscellaneous integer and floating point instructions are missing.

BreadFish64:
There's also [ryujit's](https://github.com/dotnet/coreclr/tree/master/src/jit) code gen stuff

BreadFish64:
And, while unappealing, the possibility of writing one myself

MerryMage:
Stick with dolphin for now; we require rounding mode support.

BreadFish64:
Ok, well I'll get the CI up and running on a different branch, and PR *that*

PR #360, opened 2018-08-22
Implement perfmap

MerryMage:
@degasus could you take a look?

This should produce a perfmap on linux.


src/backend/x64/emit_x64.h
```diff
@@ -85,9 +86,11 @@ class EmitX64 {
 #undef A64OPC
 
     // Helpers
+    virtual std::string LocationDescriptorToFriendlyName(const IR::LocationDescriptor&) = 0;
```
lioncash: This can likely be a const qualified virtual function.

src/backend/x64/perf_map.cpp
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    std::setvbuf(file, nullptr, _IONBF, 0);
+}
+
+void PerfMapRegister(const void* entrypoint, size_t size, const std::string& friendly_name) {
+    std::lock_guard guard{mutex};
+
+    if (!file) {
+        if (!std::getenv("PERF_BUILDID_DIR")) {
```
degasus: Please move this check into OpenFile. It can still set file = nullptr.

src/backend/x64/perf_map.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <fmt/format.h>
+
+namespace Dynarmic::BackendX64 {
+
+static std::mutex mutex;
+static std::FILE* file = nullptr;
+
+static void OpenFile() {
+    const pid_t pid = getpid();
+    const std::string filename = fmt::format("/tmp/perf-{:d}.map", pid);
```
degasus: /tmp is the default location, but it shall be overwritten if PERF_BUILDID_DIR is set.

src/backend/x64/a64_emit_x64.cpp
```diff
@@ -999,6 +998,13 @@ void A64EmitX64::EmitA64ExclusiveWriteMemory128(A64EmitContext& ctx, IR::Inst* i
     EmitExclusiveWrite(ctx, inst, 128);
 }
 
+std::string A64EmitX64::LocationDescriptorToFriendlyName(const IR::LocationDescriptor& ir_descriptor) {
+    const A64::LocationDescriptor descriptor{ir_descriptor};
+    return fmt::format("a64_a{:08X}_fpcr{:08X}",
```
degasus: is the "a" for address? However, it may fit to the hex printing of the PC afterwards, so maybe just drop this letter.

degasus:
LGTM

PR #362, opened 2018-08-29
emit_x64_vector: Remove fallbacks in EmitVectorSignExtend32() and EmitVectorSignExtend64()

lioncash:
These are fairly trivial enough that they can be done without a fallback without too much work involved.


src/backend/x64/emit_x64_vector.cpp
```diff
@@ -2465,26 +2465,45 @@ void EmitX64::EmitVectorSignExtend16(EmitContext& ctx, IR::Inst* inst) {
...
+        code.movsxd(gpr_tmp1, gpr_tmp1.cvt32());
+
+        code.pshufd(a, a, 0b11100101);
+
+        code.movd(gpr_tmp2.cvt32(), a);
+        code.movsxd(gpr_tmp2, gpr_tmp2.cvt32());
+
+        code.movq(a, gpr_tmp1);
+        code.movq(xmm_tmp, gpr_tmp2);
+        code.punpcklqdq(a, xmm_tmp);
```
MerryMage: perhaps, instead (two instructions total):

```
code.pshufd(a, a, 0b01010000);
code.psraq(a, 32);
```

lioncash: psraq (or rather vpsraq) seems to only be a thing in AVX512VL

MerryMage: Ah, my mistake. How about the following:
```
code.movabs(b, a);
code.psrad(b, 31);
code.punpckldq(a, b);
```

lioncash: Done

lioncash:
Failure not related to these changes

PR #365, opened 2018-08-31
emit_x64_vector: Provide AVX paths for EmitVector{Max, Min}{S64, U64}

lioncash:
Reppin' b-b-b-blend city


MerryMage:
`pblendvb` performs better than `vpblendvb`.
`xmm0` is always free for this purpose, and you can use `xmm0` without needing to tell the register allocator.

lioncash:
Amended

PR #373, opened 2018-09-08
A64: Implement URECPE and URSQRTE

lioncash:


lioncash:
Failure is due to the build timing out.

MerryMage:
re math_util.h:

We could expose `RecipEstimate` available in `FPRecipEstimate.cpp` and `RecipSqrtEstimate` available in `FPRSqrtEstimate.cpp`.

Do note that those functions return an 8 bit number, and you will have to set the 9th bit.

lioncash:
Ah, didn't know we already had the algorithms in place, will change this to use those.

lioncash:
(same failure as before; build timed out)

PR #375, opened 2018-09-08
emit_x64_vector: Simplify "position == 0" cases in EmitVectorExtract() and EmitVectorExtractLower()

lioncash:
In these cases we can simply just do nothing at all (in the EmitVectorExtract() case) or just only emit a `movq` (in the EmitVectorExtractLower() case), which also allows using one register less.


lioncash:
Build timed out

PR #388, opened 2018-09-16
A64: Implement SQSHL's register and immediate vector variants

lioncash:


lioncash:
Failure isn't related to the changes in this PR

PR #390, opened 2018-09-17
data_processing_bitfield: Move datasize variables after early-exit conditionals

lioncash:
Moves the declaration of datasize to the scope that it's used within. This also takes the opportunity to apply const where applicable, and make early-exits all vertically consistent with one another.


lioncash:
failure is related to FDIV and not this change.

PR #397, opened 2018-09-22
decoders: Cast to correctly-sized type before shifting

VelocityRa:
Fixes decoding for 64-bit instructions.

Edit: Does not help/apply to any currently supported ARM versions (since all are 32-bit length or below), it's for future-proofing should such an arch be supported.

### DIFF
--- a/src/frontend/decoder/decoder_detail.h
+++ b/src/frontend/decoder/decoder_detail.h
@@ -36,16 +36,17 @@ private:
      * A '1' in a bitstring indicates that a one must be present at that bit position.
      */
     static auto GetMaskAndExpect(const char* const bitstring) {
+        const auto one = static_cast<opcode_type>(1);
         opcode_type mask = 0, expect = 0;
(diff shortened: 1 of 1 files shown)


lioncash:
Do you mean instructions where the length of the instruction is 64-bit (in which case ARM has none of those), or AArch64 instructions (in which case, the description should probably be more specific).

VelocityRa:
I mean instructions where the length of the instructions is 64-bit.
I'm using dynarmic's cool decoder infrastructure for a GPU format whose instructions are that size. More specifically PlayStation Vita's `PowerVR SGX543MP4+`.

Figured I may as well push this upstream. I know that ARM has only up to 32-bit instructions, but why not be more generic?

lioncash:
> but why not be more generic?

Sure, but given this is an ARM-oriented project, you should generally include those details in the commit message to avoid confusion

lioncash:
Ah, sorry for taking so long, never got a notification for the updated commit/push. Thanks!

PR #399, opened 2018-09-25
travis: Make macOS builds use Xcode 10

lioncash:
Keeps the toolchain up to date and allows the use of things like `<optional>` and `<variant>` from the standard library.


MerryMage:
Does use of <optional> &c require a deployment target of 10.13?

(The reason I last bumped it from 10.9 to 10.12 was because `std::uncaught_exceptions` required it.)

lioncash:
Nope, seems to be fine. Misread the targeting chart provided by travis

PR #402, opened 2018-09-30
Fix Building on Android

BreadFish64:
Android studio currently ships CMake 3.6 in it's SDK, which does not support `set(CMAKE_CXX_STANDARD 17)`
The NDK also reports the architecture as ARM64 instead of Aarch64 when compiling for 64 bit ARM targets.

The PAGE_SIZE macro in Android's user.h conflicts with a variable in `a64_emit_x64.cpp`. Considering it's not a header, the simplest solution was to undefine it.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -253,7 +253,7 @@ if (ARCHITECTURE_x86_64)
     else()
         target_sources(dynarmic PRIVATE backend/x64/exception_handler_generic.cpp)
     endif()
-elseif(ARCHITECTURE_Aarch64)
+elseif(ARCHITECTURE_Aarch64 OR ARCHITECTURE_ARM64)
     target_sources(dynarmic PRIVATE
          backend/A64/emitter/a64_emitter.cpp
          backend/A64/emitter/a64_emitter.h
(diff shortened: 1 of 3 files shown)


src/CMakeLists.txt
```diff
@@ -253,7 +253,7 @@ if (ARCHITECTURE_x86_64)
     else()
         target_sources(dynarmic PRIVATE backend/x64/exception_handler_generic.cpp)
     endif()
-elseif(ARCHITECTURE_Aarch64)
+elseif(ARCHITECTURE_Aarch64 OR ARCHITECTURE_ARM64)
```
lioncash: ARCHITECTURE_ARM64 isn't defined anywhere, so this would do nothing.

jroweboy:
@lioncash yes you *can* upgrade past 3.6, but it is still considered experimental. Both Breadfish and I have tried to get it to work, and it crashes with the ever so helpful output "CMake Error." I would like to revisit it later when it becomes officially supported, but without any useful error messages we couldn't get that to work.

src/CMakeLists.txt
BreadFish64: image
... clearly it's defined somewhere

BreadFish64: image
and the files are in the project as expected

BreadFish64:
I did manage to get a newer version of cmake working and 3.10 will be the standard in Android studio 3.4 anyway, so I'm gonna go ahead and close this

PR #403, opened 2018-10-04
Add several convenience functions to the Value class

lioncash:
Referenced in #401 that I'd open a PR for other convenience functions suggested in it. Notably, a signed variant to `GetImmediateAsU64()`. This also adds three other convenience functions, which are:

- IsSignedImmediate()
- IsUnsignedImmediate()

These would be useful for testing for values without the need to always do:

```cpp
if (x.IsImmediate() && x.GetImmediateAsU64() == ...)
```

- IsZero()

Does what it says. This is added simply because zero is the most common constant to test against in arithmetic, and is almost always a no-op or has some other beneficial quality to it, so it can have its own compact shorthand.

### DIFF
--- a/src/frontend/ir/value.cpp
+++ b/src/frontend/ir/value.cpp
@@ -194,6 +194,14 @@ u64 Value::GetImmediateAsU64() const {
     }
 }
 
+bool Value::IsSignedImmediate(s64 value) const {
+    return IsImmediate() && GetImmediateAsS64() == value;
+}
+
+bool Value::IsUnsignedImmediate(u64 value) const {
+    return IsImmediate() && GetImmediateAsU64() == value;
+}
+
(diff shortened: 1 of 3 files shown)


src/frontend/ir/value.cpp
```diff
@@ -155,6 +155,10 @@ Cond Value::GetCond() const {
     return inner.imm_cond;
 }
 
+s64 Value::GetImmediateAsS64() const {
+    return static_cast<s64>(GetImmediateAsU64());
```
MerryMage: It would probably make more sense to sign extend the value (e.g. an 8 bit value would be sign-extended to 64 bits).

lioncash: Done

PR #406, opened 2018-10-08
constant_propagation_pass: Fold Mul32 and Mul64 cases where applicable

lioncash:
Multiplication operations can currently be folded if:

1. Both arguments are known constant values
2. Either operand is zero (in which case the result is also zero)
3. Either operand is one (in which case the result is the non-one operand).

We could also technically fold on the cases where:

4. `x * -1 -> -x`
5. `-1 * y -> -y`

but I'm not certain what the best way for doing that would be currently (from an API point of view)

### DIFF
--- a/src/ir_opt/constant_propagation_pass.cpp
+++ b/src/ir_opt/constant_propagation_pass.cpp
@@ -13,6 +13,16 @@
 namespace Dynarmic::Optimization {
 namespace {
 
+// Tiny helper to avoid the need to store based off the opcode
+// bit size all over the place within folding functions.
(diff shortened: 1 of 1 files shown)


MerryMage:
> but I'm not certain what the best way for doing that would be currently (from an API point of view)

In theory one would do something along these lines:

```
IR::IREmitter ir{block};
ir.SetInsertionPoint(inst);
auto neg = ir.Neg(inst->GetArg(n));
inst->ReplaceUsesWith(neg);
```

but that part of the API hasn't really been tested.

Do ensure the insertion point is correct (i.e.: no uses before definitions).

PR #408, opened 2018-10-12
constant_propagation_pass: Add 64-bit variants of shifts to the pass

lioncash:
These optimizations can also apply to the 64-bit variants of the shift opcodes; we just need to check if the instruction has an associated pseudo-op before performing the 32-bit variant's specifics.

While we're at it, we can also relocate the code to its own function like the rest of the cases to keep organization consistent.

### DIFF
--- a/src/ir_opt/constant_propagation_pass.cpp
+++ b/src/ir_opt/constant_propagation_pass.cpp
@@ -154,6 +154,26 @@ void FoldOR(IR::Inst& inst, bool is_32_bit) {
     }
 }
 
+void FoldShifts(IR::Inst& inst) {
+    IR::Inst* carry_inst = inst.GetAssociatedPseudoOperation(IR::Opcode::GetCarryFromOp);
+
+    // The 32-bit variants can contain 3 arguments, while the
+    // 64-bit variants only contain 2.
+    if (inst.NumArgs() == 3 && !carry_inst) {
+        inst.SetArg(2, IR::Value(false));
+    }
+
+    const auto shift_amount = inst.GetArg(1);
+    if (!shift_amount.IsZero()) {
+        return;
+    }
+
+    if (carry_inst) {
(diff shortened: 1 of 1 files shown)


src/ir_opt/constant_propagation_pass.cpp
```diff
@@ -154,6 +154,24 @@ void FoldOR(IR::Inst& inst, bool is_32_bit) {
...
+    if (has_pseudo_op && !inst.GetAssociatedPseudoOperation(IR::Opcode::GetCarryFromOp)) {
+        inst.SetArg(2, IR::Value(false));
+    }
+
+    const auto shift_amount = inst.GetArg(1);
+    if (!shift_amount.IsZero()) {
+        return;
+    }
+
+    IR::Inst* carry_inst = inst.GetAssociatedPseudoOperation(IR::Opcode::GetCarryFromOp);
```
MerryMage: Move this up to line 159 perhaps.

src/ir_opt/constant_propagation_pass.cpp
```diff
@@ -154,6 +154,24 @@ void FoldOR(IR::Inst& inst, bool is_32_bit) {
     }
 }
 
+void FoldShifts(IR::Inst& inst) {
+    const bool has_pseudo_op = inst.HasAssociatedPseudoOperation();
```
MerryMage: `HasAssociatedPseudoOperation` probably [doesn't do what you think it does](https://github.com/MerryMage/dynarmic/blob/master/src/frontend/ir/microinstruction.cpp#L484). Sorry.

While it would be nice to check if it is valid for the opcode to have an attached pseudoinstruction, it's probably unnecessary.

lioncash: Replaced it with an argument check instead, since we need something to differ the behavior here. 32-bit variants have 3 arguments, while the 64-bit ones only have 2. Leaving it as is will cause an assertion to trigger.

PR #412, opened 2018-10-31
Remove always-true condition

janisozaur:
fltsize is only ever set to 16 on branch that immediately returns, there
is no point in checking it again, as the condition is always true.

https://github.com/MerryMage/dynarmic/blob/52e646408fffccc26a03383adeafb0afc435dde0/src/frontend/A64/translate/impl/floating_point_conversion_integer.cpp#L82-L83


MerryMage:
While I agree in principle with this commit, once half-precision support is added, the `return` statement for that case would be removed.

janisozaur:
Fair point. Would you accept such an explanation instead then?

MerryMage:
Reasonable.

PR #413, opened 2018-10-31
Enable ninja and ccache on travis

janisozaur:


janisozaur:
On a fully loaded and hit cache, this reduced total time spent on Travis from 43 min 19 sec to 22 min 44 sec

PR #414, opened 2018-11-24
fuzz_util: Simplify result return in InstructionGenerator's Generate() function

lioncash:
This can just be a simple direct return without a separated declaration and assignment.


lioncash:
Failure isn't related to the changes

PR #418, opened 2018-11-24
constant_propagation_pass: Handle folding for Least/MostSignificant{Bit, Byte, Half, Word} opcodes

lioncash:
These are quite trivial to fold, given it just involves shifting or truncating the retrieved value.

### DIFF
--- a/src/ir_opt/constant_propagation_pass.cpp
+++ b/src/ir_opt/constant_propagation_pass.cpp
@@ -98,6 +98,51 @@ void FoldEOR(IR::Inst& inst, bool is_32_bit) {
     }
 }
 
+void FoldLeastSignificantByte(IR::Inst& inst) {
+    if (!inst.AreAllArgsImmediates()) {
+        return;
+    }
+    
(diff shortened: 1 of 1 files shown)


src/ir_opt/constant_propagation_pass.cpp
```diff
@@ -98,6 +98,51 @@ void FoldEOR(IR::Inst& inst, bool is_32_bit) {
...
+    inst.ReplaceUsesWith(IR::Value{static_cast<u32>(operand.GetImmediateAsU64())});
+}
+
+void FoldMostSignificantBit(IR::Inst& inst) {
+    if (!inst.AreAllArgsImmediates()) {
+        return;
+    }
+
+    const auto operand = inst.GetArg(0);
+    inst.ReplaceUsesWith(IR::Value{(operand.GetImmediateAsU64() >> 31) != 0});
```
VPeruS: not 63?

lioncash: No, because the opcode operates only on 32-bit values.

PR #419, opened 2018-11-24
constant_propagation_pass: Fold byte reversal opcodes where applicable

lioncash:
These are also reasonably trivial to fold away when applicable.

Specifically, if the operand is zero or all bits are set, we can just directly replace the instruction with the operand, since a swap would essentially be a no-op in these cases. In the case the operand is a different raw immediate, we can just perform the swap directly and replace the instruction with the constant value.

(force pushes were just improving the commit message)

### DIFF
--- a/src/ir_opt/constant_propagation_pass.cpp
+++ b/src/ir_opt/constant_propagation_pass.cpp
@@ -6,6 +6,7 @@
 
 #include <dynarmic/A32/config.h>
 
+#include "common/bit_util.h"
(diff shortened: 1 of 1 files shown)


MerryMage:
Special casing zero and all ones: Is this necessary; would this wrongly be handled by the rest of the code?

lioncash:
It was more to avoid doing unnecessary work. In hindsight it's kind of silly though. Will remove it.

lioncash:
Removed the unnecessary handling.

PR #421, opened 2019-01-22
tests/.../vfp_helper: Amend use of the comma operator

lioncash:
Makes the lines of code slightly nicer to read.


lioncash:
Error is related to a timeout, not this code.

PR #425, opened 2019-02-03
obvious optimization

Annomatg:
prevent requesting NumArgs multiple times
simple vector instead of a map

### DIFF
--- a/src/frontend/ir/opcodes.cpp
+++ b/src/frontend/ir/opcodes.cpp
@@ -45,34 +45,34 @@ constexpr Type NZCV = Type::NZCVFlags;
 constexpr Type Cond = Type::Cond;
 constexpr Type Table = Type::Table;
 
-static const std::map<Opcode, Meta> opcode_info {{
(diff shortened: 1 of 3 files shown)


MerryMage:
@Annomatg don't worry about it

MerryMage:
1. Indentation in this codebase uses spaces, not tabs.
2. `camel_case` for `num_args` please.
3. `const size_t num_args` as it can be `const`. (I generally prefer for things to be `const`.)
4. `static const std::vector<Meta> opcode_info`.

src/frontend/ir/opcodes.cpp
```diff
@@ -45,34 +45,34 @@ constexpr Type NZCV = Type::NZCVFlags;
...
 #undef A32OPC
 #undef A64OPC
-}};
+};
 
 } // namespace OpcodeInfo
 
 Type GetTypeOf(Opcode op) {
-    return OpcodeInfo::opcode_info.at(op).type;
+    return OpcodeInfo::opcode_info[static_cast<size_t>(op)].type;
```
lioncash: This and the below functions, should likely still retain the use of `at()` to bounds check arguments.

PR #435, opened 2019-03-02
tests/a32/testenv: Make A32TestEnv's code_mem member a std::vector

lioncash:
Makes the data member consistent with the A64 test environment.


lioncash:
Failure isn't related to this code (it's a builder timeout)

PR #437, opened 2019-03-03
A64: Implement FRECPX (single, double precision)

lioncash:


lioncash:
No idea why Github is displaying the last two commits in the wrong order (in the branch on my fork they're correctly ordered)

PR #442, opened 2019-03-04
A64: Implement scalar and vector variants of FCVTXN

lioncash:
Gotta love non-IEEE rounding mode handling.

### DIFF
--- a/src/backend/x64/emit_x64_floating_point.cpp
+++ b/src/backend/x64/emit_x64_floating_point.cpp
@@ -1034,26 +1034,46 @@ void EmitX64::EmitFPCompare64(EmitContext& ctx, IR::Inst* inst) {
 
 void EmitX64::EmitFPSingleToDouble(EmitContext& ctx, IR::Inst* inst) {
     auto args = ctx.reg_alloc.GetArgumentInfo(inst);
-    const Xbyak::Xmm result = ctx.reg_alloc.UseScratchXmm(args[0]);
+    const auto rounding_mode = static_cast<FP::RoundingMode>(args[1].GetImmediateU8());
 
-    code.cvtss2sd(result, result);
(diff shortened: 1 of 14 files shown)


src/backend/x64/emit_x64_floating_point.cpp
```diff
@@ -1046,14 +1046,24 @@ void EmitX64::EmitFPSingleToDouble(EmitContext& ctx, IR::Inst* inst) {
...
-        ForceToDefaultNaN<32>(code, result);
-    }
+    // We special-case the non-IEEE defined ToOdd rounding mode.
+    if (rounding_mode == FP::RoundingMode::ToOdd) {
+        ctx.reg_alloc.HostCall(inst, args[0]);
+        code.mov(code.ABI_PARAM2.cvt32(), ctx.FPCR());
+        code.mov(code.ABI_PARAM3.cvt32(), static_cast<u32>(rounding_mode));
+        code.lea(code.ABI_PARAM4, code.ptr[code.r15 + code.GetJitStateInfo().offsetof_fpsr_exc]);
+        code.CallFunction(&FP::FPConvert<u32, u64>);
+    } else {
```
MerryMage: This is only valid if `rounding_mode == ctx.RMode()`, given that the host rounding mode has to match the guest rounding mode.

Consider inverting the if and `if (rounding_mode == ctx.FPSCR_RMode() && rounding_mode != FP::RoundingMode::ToOdd)`

This wasn't previously a problem because we previously couldn't specify the rounding mode in v7 (this was previously intended to be controlled by `fpcr_controlled` which was always true, something that was introduced in v7 and replaced by explicit rounding modes in v8).

lioncash: Done

PR #447, opened 2019-03-06
A64: Implement CFINV, RMIF, AXFlag and XAFlag

lioncash:
Implements two of the four instructions from the ARMv8.4-CondM featureset, and all of the instructions from the ARMv8.5-CondM featureset in a naive manner.

If there's a preferable approach that doesn't make use of directly exposing of the raw values of the NZCV flag values I can change the instructions over to using that instead.

### DIFF
--- a/src/ir_opt/a64_get_set_elimination_pass.cpp
+++ b/src/ir_opt/a64_get_set_elimination_pass.cpp
@@ -23,7 +23,7 @@ void A64GetSetElimination(IR::Block& block) {
     enum class TrackingType {
         W, X,
         S, D, Q,
-        SP, NZCV,
+        SP, NZCV, NZCVRaw,
     };
     struct RegisterInfo {
         IR::Value register_value;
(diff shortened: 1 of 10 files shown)


src/ir_opt/a64_get_set_elimination_pass.cpp
```diff
@@ -35,6 +35,7 @@ void A64GetSetElimination(IR::Block& block) {
     std::array<RegisterInfo, 32> vec_info;
     RegisterInfo sp_info;
     RegisterInfo nzcv_info;
+    RegisterInfo nzcv_raw_info;
```
MerryMage: s/`nzcv_raw_info`/`nzcv_info`/g. Keep `TrackingType::NZCVRaw`.

Yeah I admit this pass is not the most intuitive code.

Here we have two different ways of setting the same underlying 'register', `SetNZCVRaw` and `SetNZCV`. So we want either of these instructions to be tracked similarly.

`SetNZCVRaw` puts a value of type of `TrackingType::NZCVRaw` into `nzcv_info`, invalidating any previous `TrackingType::NZCV` or `TrackingType::NZCVRaw` put into there.

`SetNZCV` puts a value of type of `TrackingType::NZCV` into `nzcv_info`, invalidating any previous `TrackingType::NZCV` or `TrackingType::NZCVRaw` put into there.

`GetNZCV` gets a value of type `TrackingType::NZCV` from `nzcv_info` if it's available.

PR #448, opened 2019-03-08
 A64: Implement SQSHRN, SQSHRUN, and UQSHRN's scalar variants

lioncash:
These can just be implemented in terms of the vector variants for the time being.


lioncash:
Failure seems to be related to a timeout.

PR #449, opened 2019-03-08
common/fp/info: Add specialization of FPInfo for half-precision floating point

lioncash:
Puts the necessary info struct in place for further use.

### DIFF
--- a/src/common/fp/info.h
+++ b/src/common/fp/info.h
@@ -15,6 +15,29 @@ template<typename FPT>
 struct FPInfo {};
 
 template<>
+struct FPInfo<u16> {
+    static constexpr size_t total_width = 16;
+    static constexpr size_t exponent_width = 5;
+    static constexpr size_t explicit_mantissa_width = 10;
+    static constexpr size_t mantissa_width = explicit_mantissa_width + 1;
+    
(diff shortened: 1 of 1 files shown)


lioncash:
Failure is related to a timeout.

src/common/fp/info.h
```diff
@@ -14,6 +14,29 @@ namespace Dynarmic::FP {
...
 struct FPInfo {};
 
+template<>
+struct FPInfo<u16> {
+    static constexpr size_t total_width = 16;
+    static constexpr size_t exponent_width = 5;
+    static constexpr size_t explicit_mantissa_width = 10;
+    static constexpr size_t mantissa_width = explicit_mantissa_width + 1;
+    
+    static constexpr u32 implicit_leading_bit = u32(1) << explicit_mantissa_width;
```
MerryMage: `u32` vs `u16`?

lioncash: The only reason I preferred opting for u32 here is that it preserves the unsignedness of values across operations better when used within calculations relative to the other specializations (e.g. operations on u16s promote to `int`, while the other two specializations would still result the respective unsigned type in most cases).

I can still change it over if `u16` is preferable however.

MerryMage: lol integer promotion rules

fine.

PR #453, opened 2019-03-11
backend/x64: Allow fsizes of 16 in ChooseOnFsize

lioncash:
Given half-precision instructions aren't present on x64, we can allow the size, but assert on it, given all half-precision instructions are going to have to fall back to softfloat implementations.

This allows functions that have a fast path (that uses ChooseOnFsize), and a softfloat fallback to be easily usable with half-precision floating point overloads.


MerryMage:
This might be a bit of dumb question, but if `FCODE` isn't used on a half-precision path, wouldn't this not be necessary?

lioncash:
Yeah, that could work too. I'll go for that approach instead.

PR #454, opened 2019-03-20
A64: Implement SQRDMULH and SQDMULL{2}'s vector indexed element variants

lioncash:


lioncash:
Failure is related to a timeout on the SSE3 builder

PR #455, opened 2019-03-21
A64: Implement SQRDMULH and SQDMULL's scalar indexed variants

lioncash:
These can be implemented in terms of the vector IR opcodes.


lioncash:
Failure is related to a timeout

PR #456, opened 2019-03-22
A64: Enable FMOV (general) for half-precision floating point

lioncash:
This just transfers values between vector registers and general-purpose registers with no conversions performed, so this is trivial to add support for half-precision to.


src/frontend/A64/translate/impl/floating_point_conversion_integer.cpp
```diff
@@ -77,17 +77,15 @@ bool TranslatorVisitor::FMOV_float_gen(bool sf, Imm<2> type, Imm<1> rmode_0, Imm
         }
         fltsize = 128;
         break;
-    default:
```
MerryMage: `default` is one way to not invoke the wrath of compiler warnings (`fltsize` uninitialized).

Alternatively one supposes an immediately invoked lambda expression with a way to handle the `type == 0b10 && rmode_0 != 1` case would be desireable.

lioncash: Done

PR #457, opened 2019-03-23
A64: Handle half-precision floating point in floating-point FCVT, FCVTL, and FCVTN

lioncash:
Knocks out a handful of fallbacks to the interpreter.

On top of implementing half-precision floating-point paths for the relevant instructions, this also fixes two bugs that slipped through in the FPUnpackCV and FPConvert implementations (my bad).


lioncash:
Failure is related to a timeout on the SSE3 builder.

PR #461, opened 2019-03-24
externals: Update fmt to 5.3.0

lioncash:
Keeps the formatting library up to date.

Changelog can be seen [here](https://github.com/fmtlib/fmt/releases)


MerryMage:
Merged as 89581c0c8c69acd264b2075df404350f3e632d8d.

PR #466, opened 2019-04-11
A64: Implement FCMLA's indexed element variant

lioncash:
With this, all of the instructions introduced with ARMv8.3-CompNum have an implementation.


lioncash:
Failure appears to be due to a timeout.

PR #471, opened 2019-04-12
A64: Implement SQRDMULH's scalar vector variant

lioncash:
Implements the scalar variant in terms of the vector variant for the time being.


lioncash:
Failure appears to be a timeout.

PR #477, opened 2019-04-13
A64: Handle half-precision variants of FRSQRTE

lioncash:


MerryMage:
Now conflicting.

lioncash:
Amended

PR #485, opened 2019-04-20
A32: Handle all ARM-mode arithmetic and data processing instructions added in ARMv6T2 and ARMv7

lioncash:
Now that Unicorn is in place, these can all be implemented, since we're not limited by an ARMv6K interpreter.


lioncash:
This was mainly to ensure the Windows build still built fine. Running the tests locally has them passing fine. A few unrelated kinks need to be worked out of the A32 fuzzing (particularly related to load/stores).

PR #486, opened 2019-04-26
A32: Implement barrier instructions introduced in ARMv7

lioncash:
Provides basic implementations of the barrier instruction introduced within ARMv7. Currently these simply mirror the behavior of the AArch64 equivalents.


src/frontend/A32/disassembler/disassembler_arm.cpp
```diff
@@ -76,6 +76,29 @@ class DisassemblerVisitor {
         return "<internal error>";
     }
 
+    static const char* BarriorOptionStr(Imm4 option) {
```
MerryMage: Barri`e`r

lioncash: whoops

PR #488, opened 2019-04-27
frontend: Move imm.h to the top-level directory of the frontends

lioncash:
Preparation to utilize the immediate type within the A32 backend as well, which will allow eliminating numerous type aliases like Imm4, Imm5, etc.

### DIFF
--- a/src/frontend/imm.h
+++ b/src/frontend/imm.h
@@ -11,10 +11,10 @@
 #include "common/common_types.h"
 #include "common/math_util.h"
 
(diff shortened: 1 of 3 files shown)


src/frontend/imm.h
```diff
@@ -14,7 +14,7 @@
 namespace Dynarmic::A64 {
```
MerryMage: Namespace.

MerryMage:
Location is fine with me.

PR #490, opened 2019-05-02
A32: Implement ARM-mode CRC32 instructions

lioncash:
Implements the ARM-mode variants of the CRC32 instructions introduced within ARMv8. This is also one of the instruction cases where there is UNPREDICTABLE behavior that is constrained (we must do one of the options indicated by the reference manual).

In both documented cases of constrained unpredictable behavior, we treat the instructions as undefined (which is one of the options allowed by the reference manual).


MerryMage:
We have generally just implemented constrained unpredictable behaviour to raise an unpredictable exception.

The idea being that if a library user cares sufficiently about unpredictable beahviour they should ideally implement it themselves.

lioncash:
Updated the commit to raise unpredictable behavior exceptions.

PR #497, opened 2019-05-03
A32/coprocessor: Remove boost from public interface

lioncash:
Removes a boost header from the public includes in favor of using the standard-provided std::variant. Given we use std::optional in the header already, we can transition over to the std::variant from boost::variant.

With this removal, this makes all of our dependencies internal to the library itself.

### DIFF
--- a/.travis/build-x86_64-macos/build.sh
+++ b/.travis/build-x86_64-macos/build.sh
@@ -4,7 +4,7 @@ set -e
 set -x
 set -o pipefail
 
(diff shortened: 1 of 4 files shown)


MerryMage:
macOS build failure: Requires either:

* Compatibility `<variant>` header, or
* `s/10.12/10.14/`

lioncash:
Which would you prefer in this case?

MerryMage:
Thinking about it, the latter, since the former makes less sense for an interface

lioncash:
Done.

PR #501, opened 2019-05-06
T32: Implement AND (imm)

MerryMage:
Proposed method of implementing T32 instructions without code duplication.

TODO:

* Add testing for T32.


MerryMage:
Closed. (Code incorrect as needs to handle being in IT block.)

PR #506, opened 2020-01-07
Variable and parameters optimization

Woo-ing:
1. In class "A32EmitX64": move local variable "block_ranges" to class member, avoid constructor calls too  often.
2. In class "RegAlloc": Parameters use reference call method


MerryMage:
Hey, thanks for your PR! Unfortunately, saving on the construction time of `reg_alloc` by a member variable isn't a micro-optimization I'm interested in.

PR #510, opened 2020-04-10
Set of changes to Exclusive Mem in A64 and Configuration

FernandoS27:
This set of changes do the next things:
* Allow changing the dynarmic's interface processor Id.
* Add a new wall_clock_cntpct to remove redundant code when a host timer is used instead of a cycle timer.
* Refactor Exclusive memory to using both dynarmic's OG approach and QEMU's approach, thus eliminating false negative when one cpu core does a non-exclusive write on memory marked as exclusive in another cpu core.
* Add an Exceptional Exit for callbacks that exit dynarmic execution in unconventional ways. This is normally done when a thread sleeps/pauses on an SVC and wakes up later. Since the code cache may have changed or the thread was migrated to another interface, it isn't safe to return to dynarmic.

Things to do:

- [ ] Refactor code that uses lambdas
- [ ] Implement correctly ExclusiveWrites without a monitor.

I'm leaving this code here for reviews and comments.


src/backend/x64/a64_exclusive_monitor.cpp
```diff
@@ -12,21 +12,17 @@
...
 }
 
 size_t ExclusiveMonitor::GetProcessorCount() const {
     return exclusive_addresses.size();
 }
 
-void ExclusiveMonitor::Mark(size_t processor_id, VAddr address, size_t size) {
+void ExclusiveMonitor::PreMark(size_t size) {
     ASSERT(size <= 16);
-    const VAddr masked_address = address & RESERVATION_GRANULE_MASK;
```
lioncash, line 25: If doing nothing in this function is intended, then it should be indicated why that's the case.

src/frontend/A64/translate/translate.h
```diff
@@ -27,6 +27,10 @@ struct TranslationOptions {
     /// If this is true, we define some behaviour for some instructions.
     bool define_unpredictable_behaviour = false;
 
+    /// This tells the translator a wall clock will be used, thus allowing it
+    /// to avoid writting certain unnecessary code only needed for cycle timers.
```
lioncash, line 31:
```suggestion
    /// to avoid writing certain unnecessary code only needed for cycle timers.
```

include/dynarmic/A64/exclusive_monitor.h
```diff
@@ -27,37 +30,57 @@ class ExclusiveMonitor {
...
     /// Marks a region containing [address, address+size) to be exclusive to
     /// processor processor_id.
-    void Mark(size_t processor_id, VAddr address, size_t size);
+    template <typename T, typename Function>
+    T ReadAndMark(size_t processor_id, VAddr address, Function op) {
+        const VAddr masked_address = address & RESERVATION_GRANULE_MASK;
+
+        Lock();
+        exclusive_addresses[processor_id] = masked_address;
+        T value = op();
```
lioncash, line 39:
```suggestion
        const T value = op();
```

include/dynarmic/A64/exclusive_monitor.h
```diff
@@ -27,37 +30,57 @@ class ExclusiveMonitor {
 
     /// Marks a region containing [address, address+size) to be exclusive to
     /// processor processor_id.
-    void Mark(size_t processor_id, VAddr address, size_t size);
+    template <typename T, typename Function>
+    T ReadAndMark(size_t processor_id, VAddr address, Function op) {
```
lioncash, line 34: This function should contain a static assert like:

```cpp
static_assert(std::is_trivially_copyable_v<T>);
```

ditto for `DoExclusiveOperation()`

include/dynarmic/A64/a64.h
```diff
@@ -106,6 +112,8 @@ class Jit final {
     /// Modify PSTATE
     void SetPstate(std::uint32_t value);
 
+    void ChangeProcessorId(std::size_t new_processor);
```
lioncash, line 115:
```suggestion
    void ChangeProcessorID(std::size_t new_processor);
```

src/backend/x64/a64_exclusive_monitor.cpp
FernandoS27: It just does the assert in the cpp to avoid adding the assertions in the header.

PR #513, opened 2020-04-30
Config: Use cycle counting to limit execution when enable_ticks is off.

FernandoS27:
This pull request relaxes the option `enable_ticks` to keep doing cycle counting to limit execution. This is because the current behavior of halt execution is prune to race conditions. This is a temporal measure until a better solution is found.


lioncash:
Commit message should contain the rationale and the documentation for enable_ticks should have this behavior choice documented.

FernandoS27:
Done.

lioncash:
The documentation comment for `enable_ticks` wasn't updated to indicate its behavior

MerryMage:
Superseeded.

PR #516, opened 2020-05-02
Set of changes to Exclusive Mem in A64 and Configuration

FernandoS27:
This set of changes do the next things:
* Allow changing the dynarmic's interface processor Id.
* Add a new wall_clock_cntpct to remove redundant code when a host timer is used instead of a cycle timer.
* Refactor Exclusive memory to using both dynarmic's OG approach and QEMU's approach, thus eliminating false negative when one cpu core does a non-exclusive write on memory marked as exclusive in another cpu core.
* Add an Exceptional Exit for callbacks that exit dynarmic execution in unconventional ways. This is normally done when a thread sleeps/pauses on an SVC and wakes up later. Since the code cache may have changed or the thread was migrated to another interface, it isn't safe to return to dynarmic.

### DIFF
--- a/src/backend/x64/a64_interface.cpp
+++ b/src/backend/x64/a64_interface.cpp
@@ -83,6 +83,15 @@ public:
         PerformRequestedCacheInvalidation();
     }
 
+    void ExceptionalExit() {
+        if (!conf.wall_clock_cntpct) {
+            const s64 ticks = jit_state.cycles_to_run - jit_state.cycles_remaining;
+            conf.callbacks->AddTicks(ticks);
+        }
+        PerformRequestedCacheInvalidation();
+        is_executing = false;
+    }
+
     void ClearCache() {
         invalidate_entire_cache = true;
         RequestCacheInvalidation();
@@ -311,6 +320,10 @@ void Jit::Reset() {
 
 void Jit::HaltExecution() {
     impl->HaltExecution();
+}
+
+void Jit::ExceptionalExit() {
+    impl->ExceptionalExit();
 }
 
 u64 Jit::GetSP() const {
--- a/src/backend/x64/a64_interface.cpp
+++ b/src/backend/x64/a64_interface.cpp
@@ -233,7 +233,8 @@ private:
 
         // JIT Compile
         const auto get_code = [this](u64 vaddr) { return conf.callbacks->MemoryReadCode(vaddr); };
-        IR::Block ir_block = A64::Translate(A64::LocationDescriptor{current_location}, get_code, {conf.define_unpredictable_behaviour});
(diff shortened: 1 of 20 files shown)


src/backend/x64/a64_interface.cpp
```diff
@@ -45,6 +45,7 @@ struct Jit::Impl final {
         , emitter(block_of_code, conf, jit)
     {
         ASSERT(conf.page_table_address_space_bits >= 12 && conf.page_table_address_space_bits <= 64);
+        ASSERT(conf.global_monitor != nullptr);
```
MerryMage, line 48: Please set `global_monitor` for all our tests if you put this here.

src/frontend/ir/microinstruction.cpp
```diff
@@ -457,7 +471,7 @@ bool Inst::AltersExclusiveState() const {
     return op == Opcode::A32ClearExclusive ||
            op == Opcode::A32SetExclusive   ||
            op == Opcode::A64ClearExclusive ||
-           op == Opcode::A64SetExclusive   ||
+           IsExclusiveMemoryRead()         ||
```
MerryMage, line 474: Please also add `IsExclusiveMemoryRead()` to the `IsMemoryRead()` function

src/backend/x64/a64_jitstate.h
```diff
@@ -60,6 +60,8 @@ struct A64JitState {
     static constexpr u64 RESERVATION_GRANULE_MASK = 0xFFFF'FFFF'FFFF'FFF0ull;
     u8 exclusive_state = 0;
     u64 exclusive_address = 0;
+    u64 exclusive_mem_0 = 0;
+    u64 exclusive_mem_1 = 0;
```
MerryMage, line 64: I don't think these are necessary anymore. Neither is `exclusive_address`.

src/frontend/A64/translate/impl/impl.cpp
```diff
@@ -308,7 +308,24 @@ void TranslatorVisitor::Mem(IR::U64 address, size_t bytesize, IR::AccType /*acc_
...
+    case 2:
+        return ir.ExclusiveReadMemory16(address);
+    case 4:
+        return ir.ExclusiveReadMemory32(address);
+    case 8:
+        return ir.ExclusiveReadMemory64(address);
+    case 16:
+        return ir.ExclusiveReadMemory128(address);
+    default:
+        ASSERT_MSG(false, "Invalid bytesize parameter {}", bytesize);
```
MerryMage, line 324: To avoid MSVC warning for 'reaching end of non-void function':

`ASSERT_FALSE("Invalid bytesize parameter {}", bytesize);`

PR #520, opened 2020-05-21
A32: Implement VSWP

lioncash:
A trivial one to implement, this just swaps the contents of two registers in place.


FearlessTobi:
Fuzzing seems to fail. I'm not if this is an issue with your implementation or the CI in general, but I think it would be good to rerun it.

lioncash:
It's a connection time out, so the fuzzing itself didn't directly fail. I also don't know how to restart it.

PR #522, opened 2020-06-15
A32: Implement VNEG 

lioncash:
Implements ASIMD's VNEG instruction.

Note: I'm not sure what the best way to handle the recent updates to Qemu is wrt the E bit, so I've attached the second commit so the automated tests can still execute.


MerryMage:
Regarding CPSR.E: You can probably remove this line in that case: https://github.com/MerryMage/dynarmic/blob/master/tests/A32/fuzz_arm.cpp#L143

lioncash:
Sounds good 👌

PR #528, opened 2020-06-17
A64 fastmem

vdwjeremy:
Direct access to host (virtual) address space with ```fastmem``` offset, propagated from A32.
The goal is to delegate page table computation and page boundaries detection to the hardware.

**Benchmark**
From memory intensive sequential read/write
```
    env.MemoryWrite32(0, 0x910004E7); // ADD X7, X7, 1
    env.MemoryWrite32(4, 0x91002042); // ADD X2, X2, 8
    env.MemoryWrite32(8, 0xF9400046); // LDR X6, [X2]
    env.MemoryWrite32(12, 0x91002063); // ADD X3, X3, 8
    env.MemoryWrite32(16, 0xF9000066); // STR X6, [X3]
    env.MemoryWrite32(20, 0xF9400046); // LDR X6, [X2]
    env.MemoryWrite32(24, 0xF9000066); // STR X6, [X3]
    env.MemoryWrite32(28, 0xF9400046); // LDR X6, [X2]
    env.MemoryWrite32(32, 0xF9000066); // STR X6, [X3]
    env.MemoryWrite32(36, 0xF9400046); // LDR X6, [X2]
    env.MemoryWrite32(40, 0xF9000066); // STR X6, [X3]
    env.MemoryWrite32(44, 0x17FFFFF5); // B -24
```
Elapse time on i5-4590 compared to page table is

- 70% when copying 2MB
- 71% when copying 20MB
- 78% when copying 200MB

### DIFF
--- a/include/dynarmic/A64/config.h
+++ b/include/dynarmic/A64/config.h
@@ -196,10 +196,11 @@ struct UserConfig {
     /// page boundary.
     bool only_detect_misalignment_via_page_table_on_page_boundary = false;
 
-    // Fastmem Pointer
-    // This should point to the beginning of a 4GB address space which is in arranged just like
-    // what you wish for emulated memory to be. If the host page faults on an address, the JIT
-    // will fallback to calling the MemoryRead*/MemoryWrite* callbacks.
+    /// Fastmem Pointer
+    /// This should point to the beginning of a 2^page_table_address_space_bits bytes
+    /// address space which is in arranged just like what you wish for emulated memory to
+    /// be. If the host page faults on an address, the JIT will fallback to calling the
+    /// MemoryRead*/MemoryWrite* callbacks.
     void* fastmem_pointer = nullptr;
     /// Determines if instructions that pagefault should cause recompilation of that block
     /// with fastmem disabled.
--- a/src/backend/x64/a64_emit_x64.cpp
+++ b/src/backend/x64/a64_emit_x64.cpp
@@ -75,7 +75,7 @@ A64EmitX64::BlockDescriptor A64EmitX64::Emit(IR::Block& block) {
     code.EnableWriting();
     SCOPE_EXIT { code.DisableWriting(); };
 
-    static const std::vector<HostLoc> gpr_order = [this]{
+    const std::vector<HostLoc> gpr_order = [this]{
         std::vector<HostLoc> gprs{any_gpr};
         if (conf.page_table) {
             gprs.erase(std::find(gprs.begin(), gprs.end(), HostLoc::R14));
@@ -930,7 +930,7 @@ template<std::size_t bitsize, auto callback>
 void A64EmitX64::EmitMemoryRead(A64EmitContext& ctx, IR::Inst* inst) {
     auto args = ctx.reg_alloc.GetArgumentInfo(inst);
 
-    if (!conf.page_table) {
+    if (!conf.page_table && !conf.fastmem_pointer) {
         ctx.reg_alloc.HostCall(inst, {}, args[0]);
         Devirtualize<callback>(conf.callbacks).EmitCall(code);
         return;
@@ -976,7 +976,7 @@ template<std::size_t bitsize, auto callback>
 void A64EmitX64::EmitMemoryWrite(A64EmitContext& ctx, IR::Inst* inst) {
     auto args = ctx.reg_alloc.GetArgumentInfo(inst);
 
-    if (!conf.page_table) {
+    if (!conf.page_table && !conf.fastmem_pointer) {
         ctx.reg_alloc.HostCall(nullptr, {}, args[0], args[1]);
         Devirtualize<callback>(conf.callbacks).EmitCall(code);
         return;
--- a/tests/A64/testenv.h
+++ b/tests/A64/testenv.h
@@ -125,3 +125,99 @@ public:
         return 0x10000000000 - ticks_left;
     }
 };
+
+class A64FastmemTestEnv final : public Dynarmic::A64::UserCallbacks {
+public:
+    u64 ticks_left = 0;
+    char* backing_memory = nullptr;
+
+    explicit A64FastmemTestEnv(char* addr) : backing_memory(addr) {}
+
+    template<typename T>
+    T read(u64 vaddr) {
+        T value;
+        memcpy(&value, backing_memory + vaddr, sizeof(T));
+        return value;
+    }
+    template<typename T>
+    void write(u64 vaddr, const T& value) {
+        memcpy(backing_memory + vaddr, &value, sizeof(T));
+    }
+
+    std::uint32_t MemoryReadCode(u64 vaddr) override {
+        return read<std::uint32_t>(vaddr);
+    }
+
+    std::uint8_t MemoryRead8(u64 vaddr) override {
+        return read<std::uint8_t>(vaddr);
+    }
+    std::uint16_t MemoryRead16(u64 vaddr) override {
+        return read<std::uint16_t>(vaddr);
+    }
+    std::uint32_t MemoryRead32(u64 vaddr) override {
+        return read<std::uint32_t>(vaddr);
+    }
+    std::uint64_t MemoryRead64(u64 vaddr) override {
+        return read<std::uint64_t>(vaddr);
+    }
+    Vector MemoryRead128(u64 vaddr) override {
+        return read<Vector>(vaddr);
+    }
+
+    void MemoryWrite8(u64 vaddr, std::uint8_t value) override {
+        write(vaddr, value);
+    }
+    void MemoryWrite16(u64 vaddr, std::uint16_t value) override {
+        write(vaddr, value);
+    }
+    void MemoryWrite32(u64 vaddr, std::uint32_t value) override {
+        write(vaddr, value);
+    }
+    void MemoryWrite64(u64 vaddr, std::uint64_t value) override {
+        write(vaddr, value);
+    }
+    void MemoryWrite128(u64 vaddr, Vector value) override {
+        write(vaddr, value);
+    }
+
+    bool MemoryWriteExclusive8(u64 vaddr, std::uint8_t value, [[maybe_unused]] std::uint8_t expected) override {
+        MemoryWrite8(vaddr, value);
+        return true;
+    }
+    bool MemoryWriteExclusive16(u64 vaddr, std::uint16_t value, [[maybe_unused]] std::uint16_t expected) override {
+        MemoryWrite16(vaddr, value);
+        return true;
+    }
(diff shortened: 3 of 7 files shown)


src/backend/x64/a64_emit_x64.cpp
```diff
@@ -877,51 +934,76 @@ void A64EmitX64::EmitDirectPageTableMemoryWrite(A64EmitContext& ctx, IR::Inst* i
 }
 
 void A64EmitX64::EmitA64ReadMemory8(A64EmitContext& ctx, IR::Inst* inst) {
-    if (conf.page_table) {
+    if (conf.fastmem_pointer) {
+        EmitDirectMemoryRead(ctx, inst, 8);
+    }
+    else if (conf.page_table) {
```
lioncash, line 940: Our formatting combines the braces onto one line like so:

```suggestion
    } else if (conf.page_table) {
```

ditto for all other occurrences

tests/A64/testenv.h
```diff
@@ -125,3 +125,89 @@ class A64TestEnv final : public Dynarmic::A64::UserCallbacks {
...
+        return *((uint16_t*)backing_memory + (vaddr / 2));
+    }
+    std::uint32_t MemoryRead32(u64 vaddr) override {
+        return *((uint32_t*)backing_memory + (vaddr / 4));
+    }
+    std::uint64_t MemoryRead64(u64 vaddr) override {
+        return *((uint64_t*)backing_memory + (vaddr / 8));
+    }
+    Vector MemoryRead128(u64 vaddr) override {
+        return { MemoryRead64(vaddr), MemoryRead64(vaddr + 8) };
```
lioncash, line 153:
```suggestion
        return {MemoryRead64(vaddr), MemoryRead64(vaddr + 8)};
```

tests/A64/testenv.h
```diff
@@ -125,3 +125,89 @@ class A64TestEnv final : public Dynarmic::A64::UserCallbacks {
         return 0x10000000000 - ticks_left;
     }
 };
+
+class RawMemoryTestEnv final : public Dynarmic::A64::UserCallbacks {
+public:
+    u64 ticks_left = 0;
+    char* backing_memory = nullptr;
+
+    RawMemoryTestEnv(char* addr) : backing_memory(addr) {}
```
lioncash, line 134:
```suggestion
    explicit RawMemoryTestEnv(char* addr) : backing_memory(addr) {}
```

tests/A64/testenv.h
```diff
@@ -125,3 +125,89 @@ class A64TestEnv final : public Dynarmic::A64::UserCallbacks {
...
+
+    std::uint32_t MemoryReadCode(u64 vaddr) override {
+        return *((uint32_t*)backing_memory + (vaddr / 4));
+    }
+
+    std::uint8_t MemoryRead8(u64 vaddr) override {
+        return *(backing_memory + vaddr);
```
lioncash, line 141:
```suggestion
        return static_cast<std::uint8_t>(backing_memory[vaddr]);
```

tests/A64/testenv.h
```diff
@@ -125,3 +125,89 @@ class A64TestEnv final : public Dynarmic::A64::UserCallbacks {
...
+
+class RawMemoryTestEnv final : public Dynarmic::A64::UserCallbacks {
+public:
+    u64 ticks_left = 0;
+    char* backing_memory = nullptr;
+
+    RawMemoryTestEnv(char* addr) : backing_memory(addr) {}
+
+    std::uint32_t MemoryReadCode(u64 vaddr) override {
+        return *((uint32_t*)backing_memory + (vaddr / 4));
```
lioncash, line 137: This should be using memcpy to extract the value from the memory, otherwise this is undefined behavior.

Ditto for other occurrences.

tests/A64/testenv.h
```diff
@@ -125,3 +125,89 @@ class A64TestEnv final : public Dynarmic::A64::UserCallbacks {
...
+    }
+    Vector MemoryRead128(u64 vaddr) override {
+        return { MemoryRead64(vaddr), MemoryRead64(vaddr + 8) };
+    }
+
+    void MemoryWrite8(u64 vaddr, std::uint8_t value) override {
+        backing_memory[vaddr] = value;
+    }
+    void MemoryWrite16(u64 vaddr, std::uint16_t value) override {
+        *((uint16_t*)backing_memory + (vaddr / 2)) = value;
```
lioncash, line 160: Ditto for setting as well.

MerryMage:
Please match the codestyle of the rest of the codebase.

Design comments:

1. Handling of segmentation faults is required. Users of this library intentionally map inaccessible memory into a fastmem region to trigger this behaviour. This must fallback to another method of memory access. Recompilation should be a configuration option.
2. Not all users of this library find mirroring the address space desirable behaviour. This is why it’s a configuration option.
3. You can reserve a register for the fastmem base pointer.

Misc comments:

1. Your memory read/write implementations in `RawMemoryTestEnv` are incorrect. Misaligned memory accesses need to be supported.

The above is from a brief look. I will give this a proper review when I can.

include/dynarmic/A64/config.h
```diff
@@ -171,7 +171,7 @@ struct UserConfig {
     void** page_table = nullptr;
     /// Declares how many valid address bits are there in virtual addresses.
     /// Determines the size of page_table. Valid values are between 12 and 64 inclusive.
-    /// This is only used if page_table is not nullptr.
+    /// This is only used if page_table or fastmem_pointer is not nullptr.
```
MerryMage, line 174: Incorrect documentation change.

include/dynarmic/A64/config.h
```diff
@@ -196,6 +196,11 @@ struct UserConfig {
     /// page boundary.
     bool only_detect_misalignment_via_page_table_on_page_boundary = false;
 
+    // Fastmem Pointer
+    // This should point to the beginning of a 128 bits aligned 2^page_table_address_space_bits address space which is in arranged just like
```
MerryMage, line 200: The fastmem pointer doesn't have to be 128 bits aligned.

vdwjeremy: true, although it seems to be recommended : https://software.intel.com/en-us/forums/intel-isa-extensions/topic/752392

include/dynarmic/A64/config.h
vdwjeremy: actually, page_table_address_space_bits is reused for fastmem as well, as the semantic was quite similar. However if this is misleading I can introduce a new parameter.

tests/A64/testenv.h
```diff
@@ -125,3 +125,96 @@ class A64TestEnv final : public Dynarmic::A64::UserCallbacks {
...
+    u64 ticks_left = 0;
+    char* backing_memory = nullptr;
+
+    explicit RawMemoryTestEnv(char* addr) : backing_memory(addr) {}
+
+    std::uint32_t MemoryReadCode(u64 vaddr) override {
+        return *((uint32_t*)backing_memory + (vaddr / 4));
+    }
+
+    template<typename rtype> rtype read(u64 vaddr) {
```
lioncash, line 140:
```suggestion
    template<typename T>
    T Read(u64 vaddr) {
```

tests/A64/testenv.h
lioncash: This needs to be using the read helper as well.

vdwjeremy:
> Apologies for my stricter than usual comments, but this touches a part of the codebase which is difficult to adequately test.

no worry, that's helpful :)

vdwjeremy:
> Design comments:
>     1. Handling of segmentation faults is required. Users of this library intentionally map inaccessible memory into a fastmem region to trigger this behaviour. This must fallback to another method of memory access. Recompilation should be a configuration option.
>     2. Not all users of this library find mirroring the address space desirable behaviour. This is why it’s a configuration option.
>     3. You can reserve a register for the fastmem base pointer.

1. ok, I was not sure this could be useful, in that case I will add this part
2. can you develop ?
3. will do, the register will be reserved for the full block though, do you foresee potential issues ?

MerryMage:
Just FYI, I've implemented what I've said above here: https://github.com/MerryMage/dynarmic/compare/fm

vdwjeremy:
> Just FYI, I've implemented what I've said above here: https://github.com/MerryMage/dynarmic/compare/fm

Too fast !
I was working on a similar, but it won't be compatible I'm afraid, and I didn't have the R13 register yet. Do you want to continue on branch fm ?

On the exception fallback I think we had the same, but I couldn't validate it succesfully, it goes through ```A64EmitX64::FastmemCallback``` but then crashes.
I'm testing with restrictions on windows pages such as
```
    void* vmem = VirtualAlloc(nullptr, memory_size, MEM_RESERVE, PAGE_NOACCESS);
    VirtualAlloc(vmem, 1024, MEM_COMMIT, PAGE_READWRITE); // code memory
    VirtualAlloc((char*)vmem+ 1024 * 1024, 1024 * 1024, MEM_COMMIT, PAGE_NOACCESS); // no access
```
Do you get better results ?

vdwjeremy:
rebased the PR on top of fm branch, just changed :
- added unit tests
- fixed conditions to use user callbacks
The few tests I did on fallback still can't clear the SIGSEGV when faulting on memory though

MerryMage:
Hey! Apologies for getting back to this late. LGTM as is, I'll rebase onto `mp` when I can.

MerryMage:
> can't clear the SIGSEGV

Could you clarify what you mean by this?

vdwjeremy:
OK I understood what happens.
Effectivelly the MSVC debugger intercepts the segfault, that can be resumed properly and does not appear when the debugger is not attached.
But also the test framework interferes with signals. I could get a different behaviour by playing with these flags
```
CATCH_CONFIG_NO_POSIX_SIGNALS
CATCH_INTERNAL_CONFIG_NO_POSIX_SIGNALS
CATCH_CONFIG_NO_WINDOWS_SEH
CATCH_INTERNAL_CONFIG_NO_WINDOWS_SEH
```
But it starts to be messy and I can't find a way to make a portable test out of it. For now I will leave it as is, if that's ok with you.

degasus:
Is there any way how I can help to get this merged to master as well? I can confirm that this implementation is working fine. I've seen a 15% performance increase in the CPU emulation in yuzu ingame BOTW.

MerryMage:
I can rebase this branch to a commit which is yuzu-compatible. (I've removed some yuzu-specific hacks.)

degasus:
Thanks for the offer, I already did so: https://github.com/degasus/dynarmic/commits/fm
That's how I've tested this branch. However I think in the long term, this should all go to master, shouldn't it?

PR #530, opened 2020-06-18
A32: Implement ASIMD VCEQ, VCGE, VCGT, VCLE, VCLT with zero

lioncash:
Fairly self-explanatory, we can leverage the existing IR functions for the purpose of these instructions.

In the integer case, we can just insert function pointers into an array and index it, given all comparison primitives exist already for the integer side of things.


MerryMage:
Dumb qn, but could the FP case be done in the same way?

lioncash:
It could using lambda functions like so (necessary, since parameter order differs in the final two elements):

```cpp
using FloatFn = IR::U128 (*)(ArmTranslatorVisitor&, const IR::U128&, const IR::U128&);
static constexpr std::array<FloatFn, 5> fns{
    [](ArmTranslatorVisitor& v, const auto& reg_m, const auto& zero) {
        return v.ir.FPVectorEqual(32, reg_m, zero);
    },
    [](ArmTranslatorVisitor& v, const auto& reg_m, const auto& zero) {
        return v.ir.FPVectorGreaterEqual(32, reg_m, zero);
    },
    [](ArmTranslatorVisitor& v, const auto& reg_m, const auto& zero) {
        return v.ir.FPVectorGreater(32, reg_m, zero);
    },
    [](ArmTranslatorVisitor& v, const auto& reg_m, const auto& zero) {
        return v.ir.FPVectorGreaterEqual(32, zero, reg_m);
    },
    [](ArmTranslatorVisitor& v, const auto& reg_m, const auto& zero) {
        return v.ir.FPVectorGreater(32, zero, reg_m);
    },
};

return fns[static_cast<size_t>(type)](v, reg_m, zero);
```

MerryMage:
No, it's fine. I had misread the code.

MerryMage:
Apologies, reverting this commit. My fault, I forgot that ASIMD requires StandardFPSCRValue. I will implement this when able.

PR #532, opened 2020-06-18
A32: Implement several ASIMD shift instructions

lioncash:
Implements the following ASIMD shift instructions:

- VSHR
- VSRA
- VRSHR
- VRSRA
- VSHL
- VSRI
- VSLI

The rest of the shift instructions will be implemented in another PR to avoid cluttering up this PR, which is already large

### DIFF
--- a/src/frontend/A32/translate/impl/asimd_two_regs_shift.cpp
+++ b/src/frontend/A32/translate/impl/asimd_two_regs_shift.cpp
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: 0BSD
  */
 
+#include "common/assert.h"
 #include "common/bit_util.h"
 
 #include "frontend/A32/translate/impl/translate_arm.h"
@@ -53,7 +54,7 @@ bool ShiftRight(ArmTranslatorVisitor& v, bool U, bool D, size_t imm6, size_t Vd,
 
     // Technically just a related encoding (One register and modified immediate instructions)
     if (!L && Common::Bits<3, 5>(imm6) == 0) {
-        return v.UndefinedInstruction();
+        ASSERT_FALSE();
     }
 
     const auto [esize, shift_amount] = ElementSizeAndShiftAmount(true, L, imm6);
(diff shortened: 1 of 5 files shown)


src/frontend/A32/translate/impl/asimd_two_regs_shift.cpp
```diff
@@ -0,0 +1,177 @@
...
+    return true;
+}
+
+bool ArmTranslatorVisitor::asimd_VSHL(bool D, size_t imm6, size_t Vd, bool L, bool Q, bool M, size_t Vm) {
+    if (Q && (Common::Bit<0>(Vd) || Common::Bit<0>(Vm))) {
+        return UndefinedInstruction();
+    }
+
+    // Technically just a related encoding (One register and modified immediate instructions)
+    if (!L && Common::Bits<3, 5>(imm6) == 0) {
```
MerryMage, line 162: Since this should be unreachable, I prefer to `ASSERT`.

lioncash: Wouldn't fuzzing hit this case when generating test instructions?

MerryMage: It shouldn't. If it did, there's as issue with decoding.

Looking at the decoders we have currently, looks like `asimd_VMOV_imm` will not be tested before any of the shift instructions, which would be incorrect. 

This happens because the relationship between the decoders violates [this assumption](https://github.com/MerryMage/dynarmic/blob/master/src/frontend/A32/decoder/asimd.h#L33).

We can fix this by reordering the list like so, as we do in the [A64 decoder](https://github.com/MerryMage/dynarmic/blob/master/src/frontend/A64/decoder/a64.h#L47):
```
    // Exceptions to the above rule of thumb.
    const std::set<std::string> comes_first {
        "VBIC, VMOV, VMVN, VORR (immediate)"
    };

    std::stable_partition(list.begin(), list.end(), [&](const auto& matcher) {
        return comes_first.count(matcher.GetName()) > 0;
    });
```

lioncash: Ah, OK. Thanks for pointing that out. I've added a commit that makes use of the assertions.

PR #541, opened 2020-06-20
A32: Implement ASIMD VPADDL/VPADAL

lioncash:


src/frontend/A32/translate/impl/asimd_two_regs_misc.cpp
```diff
@@ -129,6 +165,11 @@ bool ArmTranslatorVisitor::asimd_VREV(bool D, size_t sz, size_t Vd, size_t op, b
     return true;
 }
 
+bool ArmTranslatorVisitor::asimd_VPADDL(bool D, size_t sz, size_t Vd, bool op, bool Q, bool M, size_t Vm) {
+    return PairedAddOperation(*this, D, sz, Vd, op, Q, M, Vm,
+                              AccumulateBehavior::None);
```
MerryMage, line 170: Picky, but cops this be on one line?

lioncash: Done

PR #561, opened 2020-10-28
emit_x64_vector: GNFI implementation of byte-lane shifts

Wunkolo:
Optimizes our byte-lane bit-shifts into a single-instruction galois field affine transformation.

An identity 8x8 bit-matrix represented as a 64-bit integer can itself be bit-shfited which can be used to arbitrarily shift the bits of each 8-bit lane left or right. At emit-time, the 64-bit matrix constant can be generated and emitted as an embedded broadcast to allow for a fast single-instruction version of unsigned(logical) bit-shifts and possibly the signed(arithmetic) bit-shifts and rotations.

### DIFF
--- a/src/backend/x64/emit_x64_vector.cpp
+++ b/src/backend/x64/emit_x64_vector.cpp
@@ -442,7 +442,7 @@ static void ArithmeticShiftRightByte(EmitContext& ctx, BlockOfCode& code, const 
     if (code.HasAVX512_Icelake()) {
         // Do a logical shift right upon the 8x8 bit-matrix, but shift in
         // `0x80` bytes into the matrix to repeat the most significant bit.
-        const u64 zero_extend = ~(~0ull << (shift_amount)) & 0x8080808080808080;
+        const u64 zero_extend = ~(0xFFFFFFFFFFFFFFFF << shift_amount) & 0x8080808080808080;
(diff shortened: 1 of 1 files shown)


Wunkolo:
Did some unit tests over on the side on implementations for both logical _and_ arithmetic shifts. Will update this PR soon.
https://gist.github.com/Wunkolo/b715746f1599acf2c7943f9bcd2ef1fd

src/backend/x64/emit_x64_vector.cpp
```diff
@@ -439,6 +439,14 @@ void EmitX64::EmitVectorAnd(EmitContext& ctx, IR::Inst* inst) {
 }
 
 static void ArithmeticShiftRightByte(EmitContext& ctx, BlockOfCode& code, const Xbyak::Xmm& result, u8 shift_amount) {
+    if (code.HasAVX512_Icelake()) {
+        // Do a logical shift right upon the 8x8 bit-matrix, but shift in
+        // `0x80` bytes into the matrix to repeat the most significant bit.
+        const u64 zero_extend = ~(~0ull << (shift_amount)) & 0x8080808080808080;
```
lioncash, line 445:
```suggestion
        const u64 zero_extend = ~(0xFFFFFFFFFFFFFFFF << shift_amount) & 0x8080808080808080;
```

MerryMage (approved):

PR #562, opened 2020-10-28
A64 IC Instructions

emuplz:
I'm looking for a bit of feedback, so I thought I'd make a WIP pull request!

`ic ivau` expects an address argument, however `ic iallu` and `ic ialluis` invalidates all instruction caches and thus _doesn't_ expect an address argument.

So I'm not really sure how the signature should look for `InstructionCacheOperationRaised`. Should we wrap the `Vaddr` in a `std::optional`? I'm all ears! 😺 

Also, the test sucks. Shall I remove it, or is there a simpler way to test that the instructions are decoded correctly?

### DIFF
--- a/include/dynarmic/A64/config.h
+++ b/include/dynarmic/A64/config.h
@@ -69,8 +69,12 @@ enum class DataCacheOperation {
 };
 
 enum class InstructionCacheOperation {
-    // IC IVAU
+    /// IC IVAU
     InvalidateByVAToPoU,
+    /// IC IALLU
+    InvalidateAllToPoU,
+    /// IC IALLUIS
+    InvalidateAllToPoUInnerSharable
 };
 
 struct UserCallbacks {
--- a/include/dynarmic/A64/config.h
+++ b/include/dynarmic/A64/config.h
@@ -68,6 +68,11 @@ enum class DataCacheOperation {
     ZeroByVA,
 };
 
+enum class InstructionCacheOperation {
+    // IC IVAU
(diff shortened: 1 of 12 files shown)


MerryMage:
`So I'm not really sure how to the signature should look for InstructionCacheOperationRaised.`

Surprisingly I'm not actually terribly fussed about this. For simplicity you could just have it be `void(InstructionCacheOperation, VAddr)` and have the second address be undefined if it doesn't make sense.

lioncash:
What merry suggested sounds good to me. I've tried thinking of other approaches to the interface and couldn't come up with anything better

emuplz:
I think it's close to complete. I'm having some issues testing it, probably because I don't fully understand the intended behaviour!

If I ran the following code

```
mov x6, 0x1337
dc ivac, X6
```

Should I expect the callback `DataCacheOperationRaised` to be invoked with `0x1337` as the `VAddr` parameter? I get different values passed to the callback, so i'm suspecting vaddr is actually a pointer to 'host' memory.

Can anyone briefly explain what's going on here?

MerryMage (approved):
Changes look good.

There's a bug in the emitter for both `EmitA64InstructionCacheOperationRaised` and `EmitA64DataCacheOperationRaised`, thanks for pointing that out.

(Incorrect values in incorrect ABI registers, thanks to the `this` pointer.)

emuplz:
Fixed, I also provided a basic test.

I removed the final specifier from the class `TestEnv`. Might have some performance impact on tests.

PR #564, opened 2020-11-22
A32: Add ASID setting

pent0:
We emulate multi-process OS and sometimes apps force the code address, with some same state flags, but the data in the code block is different. And the block unique hash was collided.

Since clear the cache was not viable solution since the code got executed very frequently, I add ASID (address-space ID) to differentiate those code blocks. It fills the rest 5 bits of the upper location descriptor, making A32 upper location descriptor full now.

By default client side who don't care about this can still use the JIT fine. This is just extra advance methods.

I do this because I think the upper location description bits is stable now (likely no more addition to it).

**Edit:** I was thinking of should I create each JIT instance for each process instead, I think this would save some memory I hope while being valid solution.

### DIFF
--- a/src/backend/x64/a32_jitstate.cpp
+++ b/src/backend/x64/a32_jitstate.cpp
@@ -72,6 +72,8 @@ u32 A32JitState::Cpsr() const {
     return cpsr;
 }
 
+static constexpr u32 UPPER_LOCATION_DESCRIPTOR_CSPR_MASK = 0b11111111'00000011;
+
 void A32JitState::SetCpsr(u32 cpsr) {
     // NZCV flags
     cpsr_nzcv = NZCV::ToX64(cpsr);
@@ -84,7 +86,7 @@ void A32JitState::SetCpsr(u32 cpsr) {
     cpsr_ge |= Common::Bit<17>(cpsr) ? 0x0000FF00 : 0;
     cpsr_ge |= Common::Bit<16>(cpsr) ? 0x000000FF : 0;
 
-    upper_location_descriptor &= 0xFFFF0000;
+    upper_location_descriptor &= ~UPPER_LOCATION_DESCRIPTOR_CSPR_MASK;
     // E flag, T flag
     upper_location_descriptor |= Common::Bit<9>(cpsr) ? 2 : 0;
     upper_location_descriptor |= Common::Bit<5>(cpsr) ? 1 : 0;
@@ -210,6 +212,7 @@ u8 A32JitState::Asid() const {
 }
 
 void A32JitState::SetAsid(u8 ASID) {
+    upper_location_descriptor &= ~(ASID_MASK << ASID_BIT_SHIFT);
     upper_location_descriptor |= (ASID & ASID_MASK) << ASID_BIT_SHIFT;
 }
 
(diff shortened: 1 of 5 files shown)


src/backend/x64/a32_jitstate.cpp
```diff
@@ -201,4 +203,21 @@ void A32JitState::SetFpscr(u32 FPSCR) {
...
+    return (upper_location_descriptor >> ASID_BIT_SHIFT) & ASID_MASK;
+}
+
+void A32JitState::SetAsid(u8 ASID) {
+    upper_location_descriptor &= ~(ASID_MASK << ASID_BIT_SHIFT);
+    upper_location_descriptor |= (ASID & ASID_MASK) << ASID_BIT_SHIFT;
+}
+
+u8 A32JitState::MaxAsidAvailable() const {
+    return (1 << ASID_BIT_COUNT);
```
lioncash, line 220:
```suggestion
    return static_cast<u8>(1U << ASID_BIT_COUNT);
```

otherwise this is truncating an `int` implicitly.

src/backend/x64/a32_jitstate.cpp
```diff
@@ -201,4 +203,21 @@ void A32JitState::SetFpscr(u32 FPSCR) {
     }
 }
 
+constexpr u32 ASID_MASK = A32::LocationDescriptor::ASID_MASK;
+constexpr u32 ASID_BIT_SHIFT = A32::LocationDescriptor::ASID_BIT_SHIFT;
+constexpr u32 ASID_BIT_COUNT = 5;
+
+u8 A32JitState::Asid() const {
+    return (upper_location_descriptor >> ASID_BIT_SHIFT) & ASID_MASK;
```
lioncash, line 211:
```suggestion
    return static_cast<u8>((upper_location_descriptor >> ASID_BIT_SHIFT) & ASID_MASK);
```

otherwise this is truncating a `u32` implicitly.

src/frontend/A32/location_descriptor.h
```diff
@@ -42,6 +45,7 @@ class LocationDescriptor {
         fpscr = (o.Value() >> 32) & FPSCR_MODE_MASK;
         cpsr.IT(ITState{static_cast<u8>(o.Value() >> 40)});
         single_stepping = (o.Value() >> 32) & 4;
+        asid = ((o.Value() >> 32) >> ASID_BIT_SHIFT) & ASID_MASK;
```
lioncash, line 48:
```suggestion
        asid = static_cast<u8>(((o.Value() >> 32) >> ASID_BIT_SHIFT) & ASID_MASK);
```

Otherwise this truncates implicitly.

pent0:
Review addressed.

MerryMage:
We now support specifying the size of the code cache which removes the main reason for doing something like this.

pent0:
How is it relevant to this? I think you still need to clear the cache. The problem is not the cache is too small, but rather clearing it every time switching the process makes it slower. I see the hesitation on commenting on this, so I don't think you like this feature either way, sorry. I have kept it on a different fork for our purpose.

MerryMage:
There's a planned future refactor where it is planned to have per-process code caches.

pent0:
Thank you.

Vào CN, 18 thg 4, 2021 vào lúc 16:02 merry ***@***.***> đã
viết:

PR #569, opened 2021-02-01
thumb32: Implement miscellaneous category instructions

lioncash:
Begins the addition of Thumb-2 instructions and extends the Thumb-1 harness to allow for Thumb-2 instruction generation and fuzzing.


MerryMage (approved):

MerryMage:
Yes, though I feel testing move towards something more like how it's done in A32/A64.

PR #582, opened 2021-02-25
thumb32: Implement most plain binary immediate instructions

lioncash:
Implements most instructions under the plain binary immediate category for Thumb-2.

Note that SSAT/USAT currently also handle partial decodes of SSAT16/USAT16 like in the manual. Is this desirable?


MerryMage:
> Note that SSAT/USAT currently also handle partial decodes of SSAT16/USAT16 like in the manual. Is this desirable?

If in the decode table, SSAT16/USAT16 come before SSAT/USAT, the check can be turned into an `ASSERT("Invalid decode");`

lioncash:
I don't know exactly what I'm supposed to change to have what you're saying work. I moved the entries in thumb32.inc and those paths are still being hit with valid instructions.

MerryMage:
Pushed 67e954a4c

PR #583, opened 2021-02-25
thumb32: Implement STRB/STRH/STR (register)

lioncash:
Implements the register variant of the basic store instructions.


src/frontend/A32/translate/impl/thumb32_store_single_data_item.cpp
```diff
@@ -0,0 +1,50 @@
...
+    if (t == Reg::PC || m == Reg::PC) {
+        return v.UnpredictableInstruction();
+    }
+
+    const auto reg_m = v.ir.GetRegister(m);
+    const auto reg_n = v.ir.GetRegister(n);
+    const auto reg_t = v.ir.GetRegister(t);
+
+    const auto shift_amount = v.ir.Imm8(static_cast<u8>(imm2.ZeroExtend()));
+    const auto offset = v.ir.LogicalShiftLeft(reg_m, shift_amount, v.ir.GetCFlag()).result;
```
MerryMage, line 25:
```suggestion
    const auto offset = v.ir.LogicalShiftLeft(reg_m, shift_amount, v.ir.Imm1(0)).result;
```

lioncash: Done

PR #585, opened 2021-03-07
thumb32: Implement Thumb-2 LDR variants

lioncash:
Implements Thumb-2's variants of the LDR{T} instructions.

Following this PR, the halfword instructions will be implemented.

### DIFF
--- a/src/frontend/A32/translate/impl/thumb32_load_word.cpp
+++ b/src/frontend/A32/translate/impl/thumb32_load_word.cpp
@@ -6,6 +6,129 @@
 #include "frontend/A32/translate/impl/translate_thumb.h"
 
 namespace Dynarmic::A32 {
+static bool ITBlockCheck(const A32::IREmitter& ir) {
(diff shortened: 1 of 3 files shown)


src/frontend/A32/translate/impl/thumb32_load_word.cpp
```diff
@@ -6,6 +6,125 @@
...
+    const IR::U32 address = P ? offset_address
+                              : reg_n;
+    const IR::U32 data = ir.ReadMemory32(address);
+
+    if (W) {
+        ir.SetRegister(n, offset_address);
+    }
+
+    if (t == Reg::PC) {
+        ir.LoadWritePC(data);
```
MerryMage, line 57: will need `ir.UpdateUpperLocationDescriptor();` prior to LoadWritePC (updates IT state which is required for PopRSBHint/FastDispatchHint to work correctly for the T32 frontent)

lioncash: Good catch! Updated the commit

PR #590, opened 2021-03-12
Prepare work on arm backend (adding xbyak_aarch64 and update CMakeLists.txt)

vricosti:
Hi,

Sorry I changed my repository and cancel my previous pull request from it.
So I am submiting again my changes from this new one, basically I have added xbyak_aarch64 and updated cmake files.
I have also added some files inside the a64 backend folder that get the same filenames as x64 because I thought it could be better to start from existing files than create from scratch but I can remove this part).


MerryMage:
This PR is unnecessary given that we already have vixl as an AArch64 emitter.

I also note that the build files as-written are broken in any case.

PR #593, opened 2021-03-24
block_of_code: Allow Fast BMI2 paths on Zen 3

Wunkolo:
BMI2 instructions such as `pdep` and `pext` have been
known to be incredibly slow on AMD. But on Zen3
and newer, the performance of these instructions
are now much greater, but previous versions of AMD
architectures should still avoid BMI2.

[On Zen 2, pdep/pext were 300 cycles. Now on Zen 3 it is 3 cycles.](https://www.anandtech.com/show/16214/amd-zen-3-ryzen-deep-dive-review-5950x-5900x-5800x-and-5700x-tested/6)
This is a big enough improvement to allow BMI2 code to
be dispatched if available. The Zen 3 architecture is checked for
by detecting the [family of the processor](https://en.wikipedia.org/wiki/List_of_AMD_CPU_microarchitectures).

Open to suggestions on formalizing the code to look better than an apparent magic number.
Will likely have to do something similar for AMD's eventual AVX512F implementation in Zen4.


src/backend/x64/block_of_code.cpp
```diff
@@ -364,7 +364,24 @@ bool BlockOfCode::HasBMI2() const {
...
-    return DoesCpuSupport(Xbyak::util::Cpu::tBMI2) && !DoesCpuSupport(Xbyak::util::Cpu::tAMD);
+    if (DoesCpuSupport(Xbyak::util::Cpu::tBMI2))
+    {
+        // BMI2 instructions such as pdep and pext have been
+        // very slow up until Zen 3. Check for Zen 3 or newer
+        // by its Family(0x19).
+        // https://en.wikichip.org/wiki/amd/cpuid
+        if (DoesCpuSupport(Xbyak::util::Cpu::tAMD))
+        {
+            u32 data[4] = {};
```
lioncash, line 375:
```suggestion
            std::array<u32, 4> data{};
```

src/backend/x64/block_of_code.cpp
```diff
@@ -364,7 +364,24 @@ bool BlockOfCode::HasBMI2() const {
...
+            cpu_info.getCpuid(1, data);
+            const u32 family_base       = (data[0] >>  8) & 0x0F;
```
lioncash, line 377: Consider using [Bits](https://github.com/MerryMage/dynarmic/blob/master/src/common/bit_util.h#L44)

from our bit utilities, so the bits being extracted is explict. Ditto for the line below as well.

PR #594, opened 2021-03-27
emit_x64_{vector_}floating_point: NaN AVX512VL+DQ optimization

Wunkolo:
AVX512VL+AVX512DQ provides instructions such as `vfixupimmp{s,d}`/`vfixupimms{s,d}` and `vfpclassp{s,d}`/`vfpclasss{s,d}` which aid in classifying denormals, QNaNs, SNaNs, and such and concisely mask-moving/converting them into other values which greatly accelerates the way we simulate ARM's style of NaN propagation. Might need more testing(currently using `dynarmic_tests`), but this PR can sustain the dialog and review of the implementation of this code path.

https://www.felixcloutier.com/x86/vfixupimmps
https://www.officedaytime.com/simd512e/simdimg/fixup.php?f=vfixupimmps

https://www.felixcloutier.com/x86/vfpclassps
https://officedaytime.com/simd512e/simdimg/fpclass.php?f=vfpclassps


src/backend/x64/emit_x64_vector_floating_point.cpp
```diff
@@ -369,19 +383,36 @@ void EmitThreeOpVectorOperation(BlockOfCode& code, EmitContext& ctx, IR::Inst* i
...
     const Xbyak::Xmm xmm_b = ctx.reg_alloc.UseXmm(args[1]);
-    const Xbyak::Xmm nan_mask = ctx.reg_alloc.ScratchXmm();
-
-    code.movaps(nan_mask, xmm_b);
     code.movaps(result, xmm_a);
-    FCODE(cmpunordp)(nan_mask, xmm_a);
-    if constexpr (std::is_member_function_pointer_v<Function>) {
-        (code.*fn)(result, xmm_b);
+    if(code.HasAVX512VLDQ()) {
+
```
lioncash, line 388: Accidental newline

src/backend/x64/emit_x64_vector_floating_point.cpp
```diff
@@ -369,19 +400,35 @@ void EmitThreeOpVectorOperation(BlockOfCode& code, EmitContext& ctx, IR::Inst* i
...
+        // Push S/QNaN values into result
+        const Xbyak::Address push_nan = code.MConst(ptr_b, 0b0001'0001);
+
+        // Move S/QNaN from a and b to result
+        FCODE(vfixupimmp)(result, xmm_a, push_nan, u8(0));
+        FCODE(vfixupimmp)(result, xmm_b, push_nan, u8(0));
+
+        // Turn all NaNs in result into QNaNs
+        const Xbyak::Address silence_nan = code.MConst(ptr_b, 0b0010'0010);
+        FCODE(vfixupimmp)(result, result, silence_nan, u8(0));
```
MerryMage, line 419: This doesn't look like it should work. (Newly generated indeterminate NaNs need to be positive NaNs.)

Wunkolo: Based on what `vfixupimmp*` and `vfpclassp*` can do, how do you propose we use them to accelerate recreating the behavior described?
This is to follow the properties seen here correct?
https://github.com/MerryMage/dynarmic/blob/master/src/backend/x64/emit_x64_floating_point.cpp#L163-L206
Might need some more guidance

MerryMage: Yes, but see also https://github.com/MerryMage/dynarmic/blob/master/src/backend/x64/emit_x64_floating_point.cpp#L322-L330

Wunkolo:
I think I might break this one up into several different prerequisite Pull Requests so that the NaN optimizations for FMA and ThreeOp/TwoOp can have their own isolated PRs and some of the other changes like embedded-broadcast and embedded-rounding optimizations and `HasAVX512VLDQ` can be on its own thing rather than held back by this one feature.

PR #595, opened 2021-04-01
backend: x64: block_of_code: Double the total code size.

bunnei:
- The current limits are being hit in yuzu with some games (e.g. newer updates of BotW and SSBU).
- Increasing this fixes slow-downs in these games due to code being recompiled.

Credit to @degasus for finding this problem.


MerryMage:
ok
i will consider making this a config option

PR #596, opened 2021-04-01
backend/x64: Fix PerfMapRegister usages.

degasus:
Both the far code and fast_dispatch_table_lookup were missing.

TODO: BlockOfCode::ClearCache also clears the perf file, however the trampolines are not generates. This leads to missing registered code.


MerryMage (approved):

degasus:
FYI: The ClearCache bug is out of scope IMO. So is anything still missing for this PR?

PR #599, opened 2021-04-19
thumb32: Implement exclusive load instructions

lioncash:
Implements the remaining exclusive loads for ARMv7


MerryMage (dismissed):

src/frontend/A32/translate/impl/thumb32_load_store_dual.cpp
```diff
@@ -135,6 +135,56 @@ bool ThumbTranslatorVisitor::thumb32_STRD_imm_2(bool U, bool W, Reg n, Reg t, Re
     return StoreDual(*this, true, U, W, n, t, t2, imm8);
 }
 
+bool ThumbTranslatorVisitor::thumb32_LDREX(Reg n, Reg t, Imm<8> imm8) {
+    if (t == Reg::PC || n == Reg::PC) {
+        return UnpredictableInstruction();
+    }
+
+    const auto address = ir.Add(ir.GetRegister(n), ir.Imm32(imm8.ZeroExtend()));
```
MerryMage, line 143:
```suggestion
    const auto address = ir.Add(ir.GetRegister(n), ir.Imm32(imm8.ZeroExtend() << 2));
```

MerryMage (changes requested):

src/frontend/A32/translate/impl/thumb32_load_store_dual.cpp
lioncash: Good catch! Addressed it.

PR #605, opened 2021-05-10
backend/x64: Implement HostFeature

Wunkolo:
This replaces the current pattern of "`HasSSE42`,`HasAVX512_Skylake`" into a more robust `HostFeature` bit-flag system to allow rapid and parallel checks of the Host machine's capabilities and higher-level features such as `HasFastBMI2`.

Currently, while JIT-ing ARM code:
* `HasFastBMI2` and future high-level features would have to run redundantly at JIT-time when this could be checked and stored just once
https://github.com/MerryMage/dynarmic/blob/5ebe11c329e5f6d378edb8c70dcc69205d1d8b2f/src/backend/x64/block_of_code.cpp#L360-L376
* Functions like `HasAVX512_Skylake` would end up calling `BlockOfCode::DoesCpuSupport` and `Xbyak::util::Cpu::has` a bunch of times
https://github.com/MerryMage/dynarmic/blob/5ebe11c329e5f6d378edb8c70dcc69205d1d8b2f/externals/xbyak/xbyak/xbyak_util.h#L514-L517
https://github.com/MerryMage/dynarmic/blob/5ebe11c329e5f6d378edb8c70dcc69205d1d8b2f/src/backend/x64/block_of_code.cpp#L414-L420
* `Xbyak::util::Cpu::has` only checks for a non-zero, so you cannot check multiple feature-bits in parallel but rather the function must be called multiple times, once for each flag. This happens in release-mode assembly:
Code_00GlEG1f1F
image
* Despite a function only needing `AVX512F+AVX512VL`, we still ask for Icelake and Skylake levels of AVX512 features. With the current pattern, we would have to introduce new functions to check for each subset and combination of AVX512 features.

I address all of these with the implementation of `HostFeature` as a bit-mask type with support for parallel feature checking with simple bit arithmetic and since it acts as a cache where the feature only needs to be checked exactly once and quickly read from after. It also lends itself for future `HasFastBMI2` sorts of cases where a more complicated routine must be ran to determine if a feature is available and ultimately just has to set a bit rather than be ran every time in the middle of JIT-ing code.

To be forward-looking, I've also tightened up the specific instruction requirements of certain AVX512 functions to a finer granularity so that future [AMD may see the gains](https://www.techpowerup.com/279129/amd-zen-4-microarchitecture-to-support-avx-512) without requiring the host to have the entire Skylake/Icelake-level AVX512 implementations just to implement a single that really only needed AVX512F. AMD is likely to just implement AVX512F+VL+DQ+BW+BF16 and not much else(BITALG, GNFI, etc).

The enum also allows for combined features. Common feature-sets like "AVX512F+AVX512VL"(which allows using AVX512 functions on xmm registers) has been aliased to a single `AVX512_Ortho` that pre-combines these bits. It can be done to shorthand other common host-feature-sets such as `AVX+FMA `and `AVX512_Ortho + AVX512BW + AVX512DQ`.

At some point a HostFeature bitmask be provided from client-code to configure the x64 backend to disable emitting BMI or AVX code if the user wishes which would allow us to do away with handling `FastBMI2` on our end. Yuzu would be the one to provide a HostFeature mask to disable BMI2 when it knows it's on an older AMD machine, Which removes the responsibility from Dynarmic which just has to care if BMI exists.


Wunkolo:
For comparison, here is how `ArithmeticShiftRightByte` looks after this PR since https://github.com/MerryMage/dynarmic/pull/605/commits/20d9675dc2305f8c4f8582c2b0c1c47ecccea70a :
image
image

Compiled with GCC 10.2.0 with `-O3`

src/backend/x64/host_feature.h
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+#pragma once
+
+#include "common/common_types.h"
+
+namespace Dynarmic::Backend::X64 {
+
+enum class HostFeature : u64{
+    SSSE3               = 1ull <<  0,
```
lioncash, line 13: Prefer upper cased literal suffixes like `ULL`, since it prevents them from being mistaken as numerals (in particular l and 1 often look very similar in some fonts).

Wunkolo:
Just for comparison. Here is the same `ArithmeticShiftByte` assembly from above since https://github.com/MerryMage/dynarmic/pull/605/commits/87d304dd296fdbacbb087cd4d89d67c64302b616 showing the much faster `HostFeature` check in action(`mov, movabs, and, cmp` compared to a function call that tests each and every bit).

Compiled with GCC 10.2.0 with `-O3`:
image

src/backend/x64/block_of_code.h
```diff
@@ -145,22 +146,10 @@ class BlockOfCode final : public Xbyak::CodeGenerator {
...
-    bool HasLZCNT() const;
-    bool HasBMI1() const;
-    bool HasBMI2() const;
-    bool HasFastBMI2() const;
-    bool HasFMA() const;
-    bool HasAVX2() const;
-    bool HasAVX512_Skylake() const;
-    bool HasAVX512_Icelake() const;
-    bool HasAVX512_BITALG() const;
+    inline bool HasHostFeature(HostFeature feature) const
```
MerryMage, line 149: @lioncash is `inline` required here?

lioncash: Nope. Member functions defined inside a class are inline by default

Wunkolo: Yep. Removing it doesn't make a difference to assembly output. Removing it.

MerryMage (approved):
Other than minor style issues (position of braces), LGTM.

Wunkolo:
> position of braces

Fixed em up real quick 👍

PR #611, opened 2021-05-24
Remove 32-bit-only `AddWithCarry` from IR emitter

SachinVin:
Remove "Legacy" AddWithCarry used in the A32 frontend, in favor of the generic one used in the A64 frontend. Also back port A32 equivalent of SetNZCV.


MerryMage:
Thanks, this was something I was planning to do at some point. Especially once I figure out the best way to handle the NZ/NZC cases in A32.

Will do a proper review when I get time.

SachinVin:
Sure, thanks. I was only planning to deal with `ADC` and `SBC` for now, although NZC should be a simple u8 `mov` right (atleast with the current implementation)? It might be a bit more work for A64 hosts, but it should be better than what is currently being emitted.

MerryMage:
The issue with NZC is that you'll need to merge the carry from the barrel shifter with the NZ from the actual operation.

SachinVin:
Ah right, I forgot that happens.

MerryMage:
A follow-up PR for `SubWithCarry` removal would be accepted.

SachinVin:
You're welcome! I'll PR sometime soon, once I get the chance.

PR #616, opened 2021-05-29
emit_x64{_vector}_floating_point: AVX512 implementation of EmitFP{Vector}Abs{32,64}

Wunkolo:
Accomplishes floating point absolute-value without touching any memory, cutting latency down by half.


Wunkolo:
Due for more testing to ensure it handles QNaNs and Infs correctly

Wunkolo:
Currently fails to handle SNaNs.

Definition of FPAbs from armv8 documentation sets the MSB to 0(The `HaveAltFP`-path is not relevant in our case):
https://developer.arm.com/documentation/ddi0596/2020-12/Shared-Pseudocode/Shared-Functions?lang=en#impl-shared.FPAbs.1
```
bits(N) FPAbs(bits(N) op)

    assert N IN {16,32,64};
    if !UsingAArch32() && HaveAltFP() then
        FPCRType fpcr = FPCR[];
        if fpcr.AH == '1' then
            (fptype, -, -) = FPUnpack(op, fpcr, FALSE);
            if fptype IN {FPType_SNaN, FPType_QNaN} then
                return op;        // When fpcr.AH=1, sign of NaN has no consequence

    return '0' : op<N-2:0>;
```

definition of `vrange` _preserves_ SNaN
https://www.felixcloutier.com/x86/vrangeps
```
RangeSP(SRC1[31:0], SRC2[31:0], CmpOpCtl[1:0], SignSelCtl[1:0])
{
    // Check if SNAN and report IE, see also Table 5-19
    IF (SRC1=SNAN) THEN RETURN (QNAN(SRC1), set IE);
    IF (SRC2=SNAN) THEN RETURN (QNAN(SRC2), set IE);
...
```

Another idea would be to use `vfixupimm` to generate `-0.0` (`0x80000000`,`0x800000000000000`) and using a `vpandn*`
if the savings are worth it.

MerryMage:
Fails automated testing.

Example failure case (not handling denormals correctly):
```
0ea0fb6f 	fabs.2s	v15, v27

v27 = 84df4fa9113ff91b'00125528a10cd689

expected v15 = 0000000000000000'00125528210cd689
actual   v15 = 0000000000000000'00000000210cd689
```

Wunkolo:
Learned a lot of tricks with `vrange*` to use for other speedups(a `FPNeg{32,64}` optimization that I'll do after this one), but not finding anything here for FpAbs after finding out how it handles NaNs and denormals

This is where I ended up for the non-vector one, though it costs an additional zero-register _and_ 3 total(possibly uop-fused) instructions. Might not be worth it compared to just a `(a >> 1) << 1` to avoid memory. This is where I got:
```cpp
        const Xbyak::Xmm mask = ctx.reg_alloc.ScratchXmm();

        // Generate -0.0(0x80000...) without touching memory
        code.pxor(mask, mask);
        FCODE(vrangep)(mask, mask, mask, FpRangeLUT(FpRangeSelect::Min, FpRangeSign::Negative));

        ICODE(vpandn)(result, mask, result);
```

The "AVX512" implementation here might ultimately just become a single EVEX-encoded `vpand a, a, mem` instruction but with an [embedded-broadcast](https://colfaxresearch.com/skl-avx512/#sec-2-8) to reduce the original 128-bit read into just a 32-bit or 64-bit read to provide some bandwidth and space gains over the non-AVX one once I do some more perf testing related to that.

Wunkolo:
Not touching memory has its gains with the instructions taking a consistently finite amount of time rather than a memory-read-latency-bound amount of time... but I don't think it's better enough to justify over the best-case scenario memory-read solution(from L1D cache):

```
% sudo ./kernel-nanoBench.sh -unroll 500 -warm_up_count 10 -asm "pand xmm0, XMMWORD PTR [r14+0x50]" -config configs/cfg_Skylake_common.txt
RDTSC: 0.79
Instructions retired: 1.00
Core cycles: 1.00
Reference cycles: 0.78
UOPS_ISSUED.ANY: 1.00
UOPS_EXECUTED.THREAD: 2.00
UOPS_DISPATCHED_PORT.PORT_0: 0.33
UOPS_DISPATCHED_PORT.PORT_1: 0.00
UOPS_DISPATCHED_PORT.PORT_2: 0.50
UOPS_DISPATCHED_PORT.PORT_3: 0.50
UOPS_DISPATCHED_PORT.PORT_4: 0.00
UOPS_DISPATCHED_PORT.PORT_5: 0.00
UOPS_DISPATCHED_PORT.PORT_6: 0.00
UOPS_DISPATCHED_PORT.PORT_7: 0.00
BR_INST_RETIRED.ALL_BRANCHES: 0.00
BR_MISP_RETIRED.ALL_BRANCHES: 0.00
MEM_LOAD_RETIRED.L1_HIT: 1.00
MEM_LOAD_RETIRED.L1_MISS: 0.00
MEM_LOAD_RETIRED.L2_HIT: 0.00
MEM_LOAD_RETIRED.L2_MISS: 0.00
MEM_LOAD_RETIRED.L3_HIT: 0.00
MEM_LOAD_RETIRED.L3_MISS: 0.00
```

```
% sudo ./kernel-nanoBench.sh -unroll 500 -warm_up_count 10 -asm "pxor xmm0, xmm0; vrangeps xmm0, xmm0, xmm0, 0x1; vpandnd xmm1, xmm0, xmm1;" -config configs/cfg_Skylake_common.txt
RDTSC: 3.00
Instructions retired: 3.00
Core cycles: 1.63
Reference cycles: 2.94
UOPS_ISSUED.ANY: 3.00
UOPS_EXECUTED.THREAD: 2.00
UOPS_DISPATCHED_PORT.PORT_0: 0.80
UOPS_DISPATCHED_PORT.PORT_1: 0.00
UOPS_DISPATCHED_PORT.PORT_2: 0.00
UOPS_DISPATCHED_PORT.PORT_3: 0.00
UOPS_DISPATCHED_PORT.PORT_4: 0.00
UOPS_DISPATCHED_PORT.PORT_5: 0.00
UOPS_DISPATCHED_PORT.PORT_6: 0.00
UOPS_DISPATCHED_PORT.PORT_7: 0.00
BR_INST_RETIRED.ALL_BRANCHES: 0.00
BR_MISP_RETIRED.ALL_BRANCHES: 0.00
MEM_LOAD_RETIRED.L1_HIT: 0.00
MEM_LOAD_RETIRED.L1_MISS: 0.00
MEM_LOAD_RETIRED.L2_HIT: 0.00
MEM_LOAD_RETIRED.L2_MISS: 0.00
MEM_LOAD_RETIRED.L3_HIT: 0.00
MEM_LOAD_RETIRED.L3_MISS: 0.00
```

PR #619, opened 2021-06-01
emit_x64_vector_floating_point: AVX512 implementation of EmitFPVector{Min,Max}{32,64}

Wunkolo:
[Reference sheet](https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-instruction-set-reference-manual-325383.pdf) of vrange{p}{s}'s comparison behavior with +-0 and NaNs
image

Aligns pretty well with [ARM's definition](https://developer.arm.com/documentation/dui0489/c/neon-and-vfp-programming/neon-general-arithmetic-instructions/vmax--vmin--vpmax--and-vpmin?lang=en), save for some NaN handling:
> max(+0.0, -0.0) = +0.0.
> min(+0.0, -0.0) = -0.0
> If any input is a NaN, the corresponding result element is the default NaN.


Wunkolo:
Test failures seem to be related to #620

Should note that since these are AVX512 instructions, they will only actually get tested if the machine running it is AVX512-enabled, which we don't really determine on the automated build and test servers.

[Every once in a while, it does seem to run on an AVX512 enabled machine](https://github.com/MerryMage/dynarmic/pull/619/checks?check_run_id=2721379100#step:10:94)

Otherwise, something like [sde](https://software.intel.com/content/www/us/en/develop/articles/intel-software-development-emulator.html) will have to be used to test

Wunkolo:
Added some unit tests for F{MIN,MAX} to verify the implementation. Seems to pass what I've been putting them through.

src/dynarmic/backend/x64/emit_x64_vector_floating_point.cpp
```diff
@@ -958,7 +966,10 @@ static void EmitFPVectorMinMax(BlockOfCode& code, EmitContext& ctx, IR::Inst* in
             // x86-64 treats differently signed zeros as equal while ARM does not.
```
MerryMage, line 966: Comment needs to be updated stating that signed zeros are correctly handled by vrangep{s,d}

Wunkolo:
Automated testing caught some failures [here](https://github.com/MerryMage/dynarmic/pull/619/checks?check_run_id=2751288481#step:10:179). Redrafting

src/dynarmic/backend/x64/constants.h
```diff
@@ -62,8 +110,8 @@ enum class FpFixup : u8 {
     Half = 0b1011,      // 0.5
     Ninety = 0b1100,    // 90.0
     HalfPi = 0b1101,    // PI/2
-    PosMax = 0b1110,    // +{FLT_MAX,DBL_MAX}
-    NegMax = 0b1111,    // -{FLT_MAX,DBL_MAX}
+    PosMax = 0b1110,    // +{FLessThan_MAX,DBL_MAX}
```
MerryMage, line 113: This looks like an off-target search and replace :p

MerryMage (approved):

src/dynarmic/backend/x64/constants.h
Wunkolo: Oh shoot whoops

Issue #620, opened 2021-06-01
Automated testing failing at msr and vcvt.f32.f16


Wunkolo:
Across different branches(And my PRs), I'm getting fuzz-test failures in particular around the [msr](https://developer.arm.com/documentation/102374/0101/Registers-in-AArch64---system-registers) instruction and vcvt.f32.f16. Also seems to have happened on main a few times.

msr failures
https://github.com/MerryMage/dynarmic/pull/619/checks?check_run_id=2721379195#step:10:144
https://github.com/MerryMage/dynarmic/pull/619/checks?check_run_id=2721379100#step:10:146
https://github.com/MerryMage/dynarmic/pull/616/checks?check_run_id=2712829271#step:10:148

mainline instances
https://github.com/MerryMage/dynarmic/runs/2710439806#step:10:817

vcvt.f32.f16 failures
https://github.com/MerryMage/dynarmic/pull/619/checks?check_run_id=2721379161#step:10:173
https://github.com/MerryMage/dynarmic/pull/616/checks?check_run_id=2712829139#step:10:171

mainline instances
https://github.com/MerryMage/dynarmic/runs/2710422780#step:10:172

Wunkolo:
The `vcvt.f32.f16` issue seems to be an issue when processing float16 NaN values like `0xff88` turning into `0xfff10000` while unicorn is expecting `0x7fc00000`(default nan) despite the DN flag not being set though.
Unicorn might be the one default-nan-ing things when it shouldn't.

MerryMage:
> Unicorn might be the one default-nan-ing things when it shouldn't.
Indeed.

MerryMage:
msr looks like a weird unicorn bug where it thinks the instruction is 2 bytes instead of 4.

MerryMage:
`vcvt` fixed in 828959caedfac2d456a0c877fda4612e35fffc03

PR #622, opened 2021-06-05
A64/config.h: Split fastmem and page_table options.

degasus:
We might want to allocate different sizes for each of them.
e.g. for the unsafe fastmem approach without bounds checking.
Or for using the full 48bit adress range (with mirrors) by allocating our real arena as close to 1<<47 as possible.


degasus:
Can someone help me why the unittests fail? It seems to be within the A32 part, which is not touched in this PR at all.

Wunkolo:
Fuzzer has running [much more often now](https://github.com/MerryMage/dynarmic/commit/1672f907af6d55f8d7d17f49832efc8e5e67c26c) during automated testing and is making us aware of other issues: https://github.com/MerryMage/dynarmic/issues/620

MerryMage (approved):

PR #624, opened 2021-06-09
emit_x64_{vector_}floating_point: Centralize implementation of FP{Vector}{Abs,Neg}

Wunkolo:
Removes dependency on the constants at the top of some files
such as `f16_negative_zero` and `f32_non_sign_mask` in favor
of the `FPInfo` trait-type.

Also removes bypass delays by selecting between instructions
such as `pand`, `andps`, or `andpd` depending on the type
and keeps them in their respective uop domain.

See https://www.agner.org/optimize/instruction_tables.pdf for
more info on bypass delays.


src/dynarmic/common/fp/info.h
```diff
@@ -20,11 +20,11 @@ struct FPInfo<u16> {
     static constexpr size_t explicit_mantissa_width = 10;
     static constexpr size_t mantissa_width = explicit_mantissa_width + 1;
 
-    static constexpr u32 implicit_leading_bit = u32(1) << explicit_mantissa_width;
-    static constexpr u32 sign_mask = 0x8000;
-    static constexpr u32 exponent_mask = 0x7C00;
-    static constexpr u32 mantissa_mask = 0x3FF;
-    static constexpr u32 mantissa_msb = 0x200;
+    static constexpr u16 implicit_leading_bit = u16(1) << explicit_mantissa_width;
```
lioncash, line 23: The purpose of these being u32 over u16 was that they won't implicitly promote to signed ints (often not what you want when dealing with unsigned arithmetic) when arithmetic operations are potentially performed on them in expressions elsewhere.

The lack of promotion keeps it consistent with the specializations for 32-bit and 64-bit floats in that they'd all consistently provide an unsigned type regardless of the operations performed on them for consistency. Now the 16-bit float is a unique case where an int can potentially be introduced in generic contexts.

Wunkolo: Ah I get what you mean. I notice the `~` operator will promote a `u16` into an `int` which can sneak up in certain contexts.

I'll drop the commit that does this

MerryMage:
> Also removes bypass delays by selecting between instructions such as pand, andps, or andpd depending on the type and keeps them in their respective uop domain

MerryMage:
> Also removes bypass delays by selecting between instructions such as pand, andps, or andpd depending on the type and keeps them in their respective uop domain

No known processor has separate single and double fp domains. I would prefer `xorps`/`andps` to be used for all operations including 16-bit ones.

PR #626, opened 2021-06-14
emit_x64_floating_point: AVX512 implementation of EmitFPMinMaxNumeric

Wunkolo:
Going to be trying to map the NaN propagation between `vrange` and the required ARM behavior:

```
NaN requirements:
op1     op2      result
SNaN    anything op1
!SNaN   SNaN     op2
QNaN    !NaN     op2
!NaN    QNaN     op1
QNaN    QNaN     op1

vrange behavior:
op1     op2      result
SNaN    anything Quiet(op1)
!SNaN   SNaN     Quiet(op2)
QNaN    !NaN     op2
!NaN    QNaN     op1
QNaN    QNaN     op1
```
`vrange` mostly follows the ARM behavior, with the exception that it quiets SNaNs while propagating them correctly so they must be re-signaled (most-significant-bit of the mantissa needs to be flipped).


MerryMage:
Do the NaNs need to be resignalled?

Wunkolo:
This comment had made me assume the NaN must be preserved as-is

https://github.com/MerryMage/dynarmic/blob/a626a2ec63a18624e191ba1c103b08d26e65a8d9/src/dynarmic/backend/x64/emit_x64_floating_point.cpp#L488-L494

but now I notice the `orps` there further down that is always making it quiet.

https://github.com/MerryMage/dynarmic/blob/a626a2ec63a18624e191ba1c103b08d26e65a8d9/src/dynarmic/backend/x64/emit_x64_floating_point.cpp#L504-L513

a single `vrange` might be all it needs if that's the case that it consumes the NaN signal and quiets it.

Wunkolo:
Realizing this now, but this corresponds to the behavior of [2008 IEEE-754 revision](https://en.wikipedia.org/wiki/IEEE_754-2008_revision#Min_and_max) which prefers the numerical value over the NaN value.

`{min}{max}nm` and `vrange` exists for the same reason:

image

> https://developer.arm.com/documentation/100076/0100/a64-instruction-set-reference/a64-floating-point-instructions/fminnm--scalar-?lang=en

image
> http://arith22.gforge.inria.fr/slides/s1-cornea.pdf

so this might be a very concise 1:1 implementation with the exception of default-nan handling.

MerryMage (approved):

PR #630, opened 2021-06-23
A32 Vector Rounding modes VRINTN, VRINTX, VRINTA, VRINTZ, VRINTM, VRINTP

Kelebek1:
I'm not sure I did this correctly, let me know if this is all wrong. I was considering compressing them all into a single function and switching on the op to save some code, but I wasn't sure you wanted something like that.

How come the exact/inexact and fpcr_controlled states aren't output in the roundps/roundpd immediate?

### DIFF
--- a/src/dynarmic/frontend/A32/decoder/asimd.inc
+++ b/src/dynarmic/frontend/A32/decoder/asimd.inc
@@ -120,12 +120,12 @@ INST(asimd_VMOVN,           "VMOVN",                    "111100111D11zz10dddd001
 INST(asimd_VQMOVUN,         "VQMOVUN",                  "111100111D11zz10dddd001001M0mmmm") // ASIMD
 INST(asimd_VQMOVN,          "VQMOVN",                   "111100111D11zz10dddd00101oM0mmmm") // ASIMD
 INST(asimd_VSHLL_max,       "VSHLL_max",                "111100111D11zz10dddd001100M0mmmm") // ASIMD
-INST(arm_UDF,               "UNALLOCATED (VRINTN)",     "111100111-11--10----01000--0----")
-INST(arm_UDF,               "UNALLOCATED (VRINTX)",     "111100111-11--10----01001--0----")
-INST(arm_UDF,               "UNALLOCATED (VRINTA)",     "111100111-11--10----01010--0----")
-INST(arm_UDF,               "UNALLOCATED (VRINTZ)",     "111100111-11--10----01011--0----")
-INST(arm_UDF,               "UNALLOCATED (VRINTM)",     "111100111-11--10----01101--0----")
-INST(arm_UDF,               "UNALLOCATED (VRINTP)",     "111100111-11--10----01111--0----")
+INST(asimd_VRINTN,          "VRINTN",                   "111100111D11zz10dddd01000QM0mmmm") // v8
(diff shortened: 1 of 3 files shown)


Kelebek1:
Had some clang-format problems. Thought I'd point that I can't run clang-format on this project, not sure why. Trying to fix it manually for now.

Untitled-1

Wunkolo:
Ah I got this when using an older version of clang-format.
[Be sure you are using at least version `12.0.0`](https://github.com/MerryMage/dynarmic/commit/53493b20249640254d02f23f23488ea0210135d6)

src/dynarmic/frontend/A32/decoder/asimd.inc
```diff
@@ -120,12 +120,12 @@ INST(asimd_VMOVN,           "VMOVN",                    "111100111D11zz10dddd001
 INST(asimd_VQMOVUN,         "VQMOVUN",                  "111100111D11zz10dddd001001M0mmmm") // ASIMD
 INST(asimd_VQMOVN,          "VQMOVN",                   "111100111D11zz10dddd00101oM0mmmm") // ASIMD
 INST(asimd_VSHLL_max,       "VSHLL_max",                "111100111D11zz10dddd001100M0mmmm") // ASIMD
-INST(arm_UDF,               "UNALLOCATED (VRINTN)",     "111100111-11--10----01000--0----")
-INST(arm_UDF,               "UNALLOCATED (VRINTX)",     "111100111-11--10----01001--0----")
-INST(arm_UDF,               "UNALLOCATED (VRINTA)",     "111100111-11--10----01010--0----")
-INST(arm_UDF,               "UNALLOCATED (VRINTZ)",     "111100111-11--10----01011--0----")
-INST(arm_UDF,               "UNALLOCATED (VRINTM)",     "111100111-11--10----01101--0----")
-INST(arm_UDF,               "UNALLOCATED (VRINTP)",     "111100111-11--10----01111--0----")
+INST(asimd_VRINTN,          "VRINTN",                   "111100111D11zz10dddd01000QM0mmmm") // ASIMD
```
MerryMage, line 123:
```suggestion
INST(asimd_VRINTN,          "VRINTN",                   "111100111D11zz10dddd01000QM0mmmm") // v8
```
These instructons were introduced in ARMv8 as far as I can tell, not in the ASIMD update. (This information will be used for instruction versioning in the future.)

MerryMage:
> I was considering compressing them all into a single function and switching on the op to save some code
You can do that if you wish.

> How come the exact/inexact and fpcr_controlled states aren't output in the roundps/roundpd immediate?

Exact should already be accounted for.
`fpcr_controlled` isn't an argument because there wasn't a need for it before, and should be added if required.

Kelebek1:
For this path https://github.com/MerryMage/dynarmic/blob/master/src/dynarmic/backend/x64/emit_x64_vector_floating_point.cpp#L1425 it just passes the rounding mode. Since that path requires `!exact`, according to the docs (https://www.felixcloutier.com/x86/roundpd#fig-4-24) bit 3 should be set to 1 for inexact, and bit 2 for the fpcr_controlled state. Am I understanding that wrongly?

MerryMage:
Huh, good point.

Kelebek1:
What can I do to resolve these failing tests? I can't see any mistakes in what I did, and the code given doesn't have any round instructions in at least.

PR #631, opened 2021-06-25
IR: Add AndNot IR instructions

Wunkolo:
`And(a, Not(b))` is a common enough IR-operation that this can
be fused into a single `AndNot` operation. On x64 this is also
a single `pandn` or `andn`(BMI1) instruction rather than two.

This PR implements `AndNot` IR opcodes for 32/64-bit registers
as well as for Vector registers. This also includes a BMI1-accelerated path
that emits a single `andn` in the case that both operands are registers.


src/dynarmic/ir/opcodes.inc
```diff
@@ -1,720 +1,723 @@
 //     opcode name,                                         return type,    arg1 type,      arg2 type,      arg3 type,      arg4 type,      ...
 
-OPCODE(Void,                                                Void,                                                                           )
-OPCODE(Identity,                                            Opaque,         Opaque                                                          )
-OPCODE(Breakpoint,                                          Void,                                                                           )
-OPCODE(CallHostFunction,                                    Void,           U64,            Opaque,         Opaque,         Opaque          )
+OPCODE(Void, Void, )
```
lioncash, line 3: The formatting opcode tabling here is pretty mangled throughout the whole file.

src/dynarmic/backend/x64/emit_x64_data_processing.cpp
```diff
@@ -1283,6 +1283,70 @@ void EmitX64::EmitAnd64(EmitContext& ctx, IR::Inst* inst) {
...
+        Xbyak::Reg32 op_b = ctx.reg_alloc.UseGpr(args[1]).cvt32();
+        code.andn(op_a, op_b, op_a);
+        ctx.reg_alloc.DefineValue(inst, op_a);
+        return;
+    }
+
+    Xbyak::Reg32 result;
+    if (args[1].IsImmediate()) {
+        result = ctx.reg_alloc.ScratchGpr().cvt32();
+        code.mov(result, u32(~args[1].GetImmediateU32()));
```
lioncash, line 1300: This cast doesn't seem necessary, since the not is acting on a u32.

src/dynarmic/ir/opcodes.inc
Wunkolo: Oh wow you're right. Clang format tore through this file.
I'll add a `// clang-format off` for this file in a separate commit

src/dynarmic/backend/x64/emit_x64_data_processing.cpp
Wunkolo: Ah I'm getting [C4319](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4319?view=msvc-160) built warnings/errors without this cast since `~` seems to up-cast the `u32` to `u64`
https://ci.appveyor.com/project/MerryMage/dynarmic/builds/39754188#L496

lioncash (approved):

lioncash:
Oh, tests seem to be failing on one buildbot, though I'm not sure it's related

Wunkolo:
> Oh, tests seem to be failing on one buildbot, though I'm not sure it's related

Seems unrelated since it's a floating point exception.

```
1: ../tests/A32/fuzz_arm.cpp:469: FAILED:
1:   {Unknown expression after the reported line}
1: due to a fatal error condition:
1:   SIGFPE - Floating point error signal
```

Wunkolo:
Sorry about the close/reopen
Github-Android app 😒

src/dynarmic/backend/x64/emit_x64_data_processing.cpp
```diff
@@ -1283,6 +1283,70 @@ void EmitX64::EmitAnd64(EmitContext& ctx, IR::Inst* inst) {
     ctx.reg_alloc.DefineValue(inst, result);
 }
 
+void EmitX64::EmitAndNot32(EmitContext& ctx, IR::Inst* inst) {
+    auto args = ctx.reg_alloc.GetArgumentInfo(inst);
+
+    if (!args[0].IsImmediate() && !args[1].IsImmediate() && code.HasHostFeature(HostFeature::BMI1)) {
+        Xbyak::Reg32 op_a = ctx.reg_alloc.UseScratchGpr(args[0]).cvt32();
+        Xbyak::Reg32 op_b = ctx.reg_alloc.UseGpr(args[1]).cvt32();
+        code.andn(op_a, op_b, op_a);
```
MerryMage, line 1292: For instructions like this I would rather than the result and operands be in separate registers (to avoid unnecessary moves by the register allocator in the UseScratch case).

A planned regalloc rewrite will assign these the same host register if this is the last use of this value.

src/dynarmic/backend/x64/emit_x64_data_processing.cpp
```diff
@@ -1283,6 +1283,72 @@ void EmitX64::EmitAnd64(EmitContext& ctx, IR::Inst* inst) {
     ctx.reg_alloc.DefineValue(inst, result);
 }
 
+void EmitX64::EmitAndNot32(EmitContext& ctx, IR::Inst* inst) {
+    auto args = ctx.reg_alloc.GetArgumentInfo(inst);
+
+    if (!args[0].IsImmediate() && !args[1].IsImmediate() && code.HasHostFeature(HostFeature::BMI1)) {
+        Xbyak::Reg32 op_a = ctx.reg_alloc.UseScratchGpr(args[0]).cvt32();
```
MerryMage, line 1290: As we're not modifying this value this can be a `UseGpr`

PR #636, opened 2021-07-27
CMakeLists: Add install rules

abouvier:
This project cannot be easily packaged without these install rules.


MerryMage:
I don't understand the need to change the include directory structure.

abouvier:
What do you mean "change"? :p
It's packaging related, using `cmake --install`: https://aur.archlinux.org/packages/dynarmic-git/
I need to copy `frontend/A32/translate/translate_callbacks.h` because this file is included by `interface/A32/config.h` (and `common/common_types.h` is include by `translate_callbacks.h`). This is weird because these two files do not seem to be part of the interface.

abouvier:
I also added CMake config files needed to use `find_package(dynarmic)` in other projects.

Tachi107:
I believe that this PR can be closed, as #652 has been merged

PR #640, opened 2021-08-08
catch2: Correct catch2 include directory

abouvier:
It's the same problem that in https://github.com/MerryMage/dynarmic/pull/639.


MerryMage:
Apologies but this is an easier correction since we don't subtree catch, fixed in 2bc86209.

PR #647, opened 2021-09-23
x64 Interface: Make Invalidation asynchronous.

FernandoS27:
This is very important in multicore systems. Its considered the users responsability that the code is no longer being ran when invalidated from another thread.


src/dynarmic/backend/x64/a64_interface.cpp
```diff
@@ -90,15 +95,21 @@ struct Jit::Impl final {
     }
 
     void ClearCache() {
+        std::unique_lock lock(invalidation_mutex);
```
merryhime, line 98:
```suggestion
        std::unique_lock lock{invalidation_mutex};
```

merryhime (approved):

src/dynarmic/backend/x64/a64_interface.cpp
```diff
@@ -90,15 +95,21 @@ struct Jit::Impl final {
...
         invalidate_entire_cache = true;
-        RequestCacheInvalidation();
+        if (is_executing) {
+            jit_state.halt_requested = true;
+        }
     }
 
     void InvalidateCacheRange(u64 start_address, size_t length) {
+        std::unique_lock lock(invalidation_mutex);
```
merryhime, line 106:
```suggestion
        std::unique_lock lock{invalidation_mutex};
```

src/dynarmic/backend/x64/a64_interface.cpp
```diff
@@ -268,6 +279,7 @@ struct Jit::Impl final {
     }
 
     void PerformRequestedCacheInvalidation() {
+        std::unique_lock lock(invalidation_mutex);
```
merryhime, line 282:
```suggestion
        std::unique_lock lock{invalidation_mutex};
```

PR #652, opened 2021-10-12
build(cmake): add install target

Tachi107:
This makes dynarmic installable, and also adds a CMake package config file, that allows projects to use `find_package(dynarmic)` to import the library.

I know #636 adds the same thing, but while experimenting with the different install options in https://github.com/merryhime/dynarmic/pull/636#discussion_r725656034 I ended up with a working patch, so I'm proposing this as well. This implements solution 2.


CMakeLists.txt
```diff
@@ -171,3 +171,32 @@ add_subdirectory(src/dynarmic)
...
+# Install
+#
+if (MASTER_PROJECT)
+    include(GNUInstallDirs)
+    include(CMakePackageConfigHelpers)
+
+    install(TARGETS dynarmic EXPORT dynarmicTargets)
+    install(EXPORT dynarmicTargets
+        NAMESPACE dynarmic::
+        DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake"
```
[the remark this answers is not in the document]
Tachi107: Ops, you're right. Thanks for spotting this :)

IvarWithoutBones (approved):
This is really useful for packaging, great work!

CMakeLists.txt
```diff
@@ -171,3 +171,32 @@ add_subdirectory(src/dynarmic)
 if (DYNARMIC_TESTS)
     add_subdirectory(tests)
 endif()
+
+#
+# Install
+#
+if (MASTER_PROJECT)
+    include(GNUInstallDirs)
```
abouvier, line 179: Since the `GNUInstallDirs` provided variable `CMAKE_INSTALL_INCLUDEDIR` is used unconditionally in `src/dynarmic/CMakeLists.txt `, this include should be placed at the top of this `CMakeLists.txt` file.

abouvier: But `src/dynarmic` subdirectory is added even before `GNUInstallDirs` is included, so the variable will not be defined anyway.

PR #653, opened 2021-11-25
decoder/arm: Make CLZ required encoding bits more relaxed

pent0:
[The official document](https://developer.arm.com/documentation/ddi0406/c/Application-Level-Architecture/Instruction-Details/Alphabetical-list-of-instructions/CLZ) marks these now ignored bits (using brackets) as should be there, but not necessary. I was encountering a JIT generate this instruction running fine on a real CPU.

image1
image2


merryhime:
Bits 8-11 and 16-19 are SBO

pent0:
This also applies to QADD, QSUB, QDADD, QDSUB, I will keep it to myself for now (encountered it).

pent0:
I think It would be better if I modify this in a fork for my need. Dynarmic should stick to normal standard, and I think trying to get the tests to pass maybe a hassle too!

PR #654, opened 2021-11-25
Add TLB feature

pent0:
Emulators that does not store 4MB page table per process (sometimes during memory constraint) will have difficulty using the page table method.

So this simple TLB struct should help in those cases, in addition, it also tracks read/write permission of guest memory at access, avoiding access violation raised to host. 

I'm not sure my x64 code written is optimized, so feel free to correct it. It should not affect core structures of Dynarmic.

Putting as draft so Merry can decide on reject/accept when available, already been used in EKA2L1 so works well. I will keep maintaining this code in the future.


src/dynarmic/interface/tlb.h
```diff
@@ -0,0 +1,122 @@
...
+
+struct TLBEntry {
+    VAddr read_addr;
+    VAddr write_addr;
+    VAddr execute_addr;
+
+    std::uint8_t *host_base;
+
+    explicit TLBEntry()
+        : read_addr(0)
```
lioncash, line 30: These can be assigned directly to the member variables

src/dynarmic/interface/tlb.h
```diff
@@ -0,0 +1,122 @@
...
+
+#include <cstdint>
+#include <cstring>
+#include <vector>
+
+namespace Dynarmic {
+
+using VAddr = std::uint64_t;
+
+enum MemoryPermission {
```
lioncash, line 16: Prefer `enum class` with helper methods as opposed to placing identifiers in the surrounding scope

src/dynarmic/interface/tlb.h
```diff
@@ -0,0 +1,122 @@
...
+    MemoryPermissionRead = 1 << 0,
```
lioncash, line 17: With an enum class, these can be shortened to drop the `MemoryPermission` prefix

src/dynarmic/interface/tlb.h
```diff
@@ -0,0 +1,122 @@
...
+    static constexpr std::uint32_t TLB_ENTRY_COUNT = 1 << TLB_BIT_COUNT;
+    static constexpr std::uint32_t TLB_BIT_MASK = (1 << TLB_BIT_COUNT) - 1;
+
+    TLBEntry entries[TLB_ENTRY_COUNT];
+
+    std::size_t page_bits;
+    std::size_t page_mask;
+
+    explicit TLB(std::size_t page_bits)
+        : page_bits(page_bits) {
```
lioncash, line 49: The assignment can be moved up into the initializer list
```suggestion
        : page_bits(page_bits), page_mask((size_t{1} << page_bits) - 1) {
```

src/dynarmic/interface/tlb.h
```diff
@@ -0,0 +1,122 @@
...
+
+    explicit TLBEntry()
+        : read_addr(0)
+        , write_addr(0)
+        , execute_addr(0)
+        , host_base(nullptr) {
+    }
+};
+
+template <std::uint32_t TLB_BIT_COUNT>
```
lioncash, line 37:
```suggestion
template <size_t TLB_BIT_COUNT>
```

src/dynarmic/interface/tlb.h
```diff
@@ -0,0 +1,122 @@
...
+struct TLB {
+public:
+    static constexpr std::uint32_t TLB_ENTRY_COUNT = 1 << TLB_BIT_COUNT;
+    static constexpr std::uint32_t TLB_BIT_MASK = (1 << TLB_BIT_COUNT) - 1;
+
+    TLBEntry entries[TLB_ENTRY_COUNT];
```
lioncash, line 43:
```suggestion
    std::array<TLBEntry, TLB_ENTRY_COUNT> entries;
```

src/dynarmic/interface/tlb.h
```diff
@@ -0,0 +1,122 @@
...
+
+    explicit TLB(std::size_t page_bits)
+        : page_bits(page_bits) {
+        page_mask = (1 << page_bits) - 1;
+        Flush();
+    }
+
+    void Flush() {
+        // Using memfill to speed up this process
+        std::memset(entries, 0, sizeof(TLBEntry) * TLB_ENTRY_COUNT);
```
lioncash, line 56: With std::array, this can be shortened to:

```cpp
entries.fill(TLBEntry{});
```

src/dynarmic/backend/x64/a32_emit_x64.cpp
```diff
@@ -959,6 +959,60 @@ void EmitDetectMisaignedVAddr(BlockOfCode& code, A32EmitContext& ctx, size_t bit
     code.SwitchToNearCode();
 }
 
+Xbyak::RegExp EmitTLBLookup(BlockOfCode& code, A32EmitContext& ctx, size_t bitsize, Xbyak::Label& abort, Xbyak::Reg64 vaddr,
+    MemoryPermission access_type) {
```
lioncash, line 963: We generally align parametters after the first `(`

merryhime:
Thought about this and will integrate with current code base when I get time.

PR #655, opened 2021-12-14
general: Rename files with duplicate names

Morph1984:
In MSVC, having files with identical filenames will result into massive slowdowns when compiling.
The approach I have taken to resolve this is renaming the identically named files in frontend/(A32, A64) to (a32, a64)_filename.cpp/h

On my system, this reduces compile times by a minute.


lioncash:
Any reason this is still a draft?

Morph1984:
I have not run clang-format on all the files yet.

merryhime (commented):
Seems reasonable. I understand why crc32.cpp is renamed to a32_crc32.cpp.

PR #656, opened 2021-12-23
constant_pool: Use `tsl::robin_map` rather than `map`

Wunkolo:
`map` is an ordinal structure with log(n) time searches.
`unordered_map` uses O(1) average-time searches and O(n) in the worst
case where a bucket has a to a colliding hash and has to start chaining.
The unordered version should speed up our general-case when looking up
constants.

I've added a trivial order-dependent(_(0,1) and (1,0) will return a
different hash_) hash to combine a 128-bit constant into a
64-bit hash that generally will not collide, using a bit-rotate to
preserve entropy.

```
[master] % hyperfine -r 100 "./dynarmic_tests --durations yes"
Benchmark 1: ./dynarmic_tests --durations yes
  Time (mean ± σ):     567.0 ms ±   6.9 ms    [User: 513.1 ms, System: 53.2 ms]
  Range (min … max):   554.4 ms … 588.1 ms    100 runs

[opt_const_pool] % hyperfine -r 100 "./dynarmic_tests --durations yes"
Benchmark 1: ./dynarmic_tests --durations yes
  Time (mean ± σ):     561.1 ms ±   4.5 ms    [User: 508.1 ms, System: 52.3 ms]
  Range (min … max):   552.6 ms … 574.2 ms    100 runs
```


src/dynarmic/backend/x64/constant_pool.h
```diff
@@ -29,7 +30,13 @@ class ConstantPool final {
 private:
     static constexpr size_t align_size = 16;  // bytes
 
-    std::map<std::tuple<u64, u64>, void*> constant_info;
+    struct ConstantHash {
+        std::size_t operator()(const std::tuple<u64, u64>& constant) const {
```
lioncash, line 34:
```suggestion
        std::size_t operator()(const std::tuple<u64, u64>& constant) const noexcept {
```

since hashing functions should never throw

src/dynarmic/backend/x64/constant_pool.h
```diff
@@ -29,7 +30,13 @@ class ConstantPool final {
...
+            return std::get<0>(constant) ^ std::rotl<u64>(std::get<1>(constant), 1);
+        }
+    };
+
+    std::unordered_map<std::tuple<u64, u64>, void*, ConstantHash> constant_info;
```
lioncash, line 39: While the code is being touched, perhaps this should be lowered to being a std::pair instead of a std::tuple to reduce instantiation overhead?

Wunkolo: Sounds good, looks like pair also guarantees a standard layout.

merryhime:
Since we're going to microoptimize this, I should note `std::unordered_map` doesn't have great performance compared to `tsl::robin_map`, which you can pull in with `#include <tsl/robin_map.h>` (already in this repository). Has basically the same API.

Wunkolo:
> `std::unordered_map` doesn't have great performance compared to `tsl::robin_map`
Yep, finding much better gains with `robin_map`:
```
[opt_const_pool] % hyperfine -r 100 "./dynarmic_tests --durations yes"
Benchmark 1: ./dynarmic_tests --durations yes
  Time (mean ± σ):     553.5 ms ±   5.6 ms    [User: 500.7 ms, System: 52.1 ms]
  Range (min … max):   545.7 ms … 569.3 ms    100 runs
```

Wunkolo:
Guess the ubuntu environment is not ready for `std::span` yet. I'll hold it off.

merryhime (approved):

PR #662, opened 2022-02-04
exclusive_monitor: Add missing call to Unlock(), and yield thread instead of busy waiting

ameerj:
This was uncovered by profiling `Pokemon Legends Arceus` in yuzu Emulator.

Surprisingly, the missing call to `Unlock()` in `CheckAndClear()` did not cause deadlock, it did, however, create extra contention over the monitor's lock.

Adding the call to `Unlock()` helped with the performance in the game and reduced the CPU time spent in the `Lock()` method, with further improvements seen by yielding the thread instead of busy waiting. 

Please let me know if any of this was specifically designed to behave this way.

Here is the profiling data from  `Pokemon Legends Arceus` used to validate the changes. These were taken at roughly the same instance in-game, at the intro to Jubilife village, within run-to-run variance

No Unlock() |  Unlock() | Unlock() + yield
:-------------------------:|:-------------------------:|:-------------------------:
no_unlock | unlock |  unlock_yield
16.35% CPU usage, ~35 fps average | 9.17% CPU usage, ~40 fps average | 8.54% CPU usage, ~42 fps average

### DIFF
--- a/src/dynarmic/backend/x64/exclusive_monitor.cpp
+++ b/src/dynarmic/backend/x64/exclusive_monitor.cpp
@@ -42,6 +42,7 @@ bool ExclusiveMonitor::CheckAndClear(size_t processor_id, VAddr address) {
             other_address = INVALID_EXCLUSIVE_ADDRESS;
         }
     }
+    Unlock();
     return true;
 }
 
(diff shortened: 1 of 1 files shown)


merryhime:
Adding `Unlock()` there is incorrect.

merryhime (changes requested):
as above

merryhime:
What would result in better performance here rather than microoptimizing these calls, would be inlining exclusive monitor access into JITted code, which would greatly reduce the call overhead.

merryhime:
close for now

PR #664, opened 2022-02-26
Inline exclusive memory access operations

merryhime:
This emulates exclusive instructions with qemu/unicorn semantics, relying purely on cmpxchg. Note that this will obviously suffer from the [ABA problem](https://en.wikipedia.org/wiki/ABA_problem); however no user of this library uses the exclusive callbacks correctly in order to avoid this issue anyway.

### DIFF
--- a/src/dynarmic/backend/x64/a32_emit_x64_memory.cpp
+++ b/src/dynarmic/backend/x64/a32_emit_x64_memory.cpp
@@ -84,6 +84,14 @@ void A32EmitX64::GenFastmemFallbacks() {
             }
         }
     }
+
+    for (int status_idx : idxes) {
+        code.align();
+        exclusive_write_fallbacks[status_idx] = code.getCurr<void (*)()>();
+        code.mov(Xbyak::Reg32{status_idx}, 1);
+        code.ret();
+        PerfMapRegister(exclusive_write_fallbacks[status_idx], code.getCurr(), fmt::format("a64_exclusive_write_fallback"));
+    }
 }
 
 std::optional<A32EmitX64::DoNotFastmemMarker> A32EmitX64::ShouldFastmem(A32EmitContext& ctx, IR::Inst* inst) const {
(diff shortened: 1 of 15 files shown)


FernandoS27:
hmn could you add the 128b version too? i think that uses cmpxchg16 and you need 64 bits to have it either way.

Other than that LGTM

merryhime:
@FernandoS27 128 bit accesses already implemented

FernandoS27:
> @FernandoS27 128 bit accesses already implemented

all good then.

src/dynarmic/backend/x64/a32_emit_x64_memory.cpp
```diff
@@ -415,40 +426,186 @@ void A32EmitX64::ExclusiveWriteMemory(A32EmitContext& ctx, IR::Inst* inst) {
...
+            std::nullopt,
+        });
+
+    code.L(end);
+
+    code.mov(code.byte[r15 + offsetof(A32JitState, exclusive_state)], u8(1));
+    code.mov(tmp, Common::BitCast<u64>(GetExclusiveMonitorValuePointer(conf.global_monitor, conf.processor_id)));
+    EmitWriteMemoryMov<bitsize>(code, tmp, value);
+    code.mov(tmp, Common::BitCast<u64>(GetExclusiveMonitorAddressPointer(conf.global_monitor, conf.processor_id)));
+    code.mov(qword[tmp], vaddr);
```
FernandoS27, line 466: This memory is volatile and will race with a real monitor ran in the host. 
https://github.com/merryhime/dynarmic/blob/5f96ca00b109d99c2ea1d4d141b3e3948edefbea/src/dynarmic/backend/x64/exclusive_monitor.cpp#L42

 Best is to add a separate storage that will be used by guest/unsafe option and should only be cleared here.
https://github.com/merryhime/dynarmic/blob/5f96ca00b109d99c2ea1d4d141b3e3948edefbea/src/dynarmic/backend/x64/exclusive_monitor.cpp#L56

src/dynarmic/backend/x64/a32_emit_x64_memory.cpp
```diff
@@ -415,40 +426,186 @@ void A32EmitX64::ExclusiveWriteMemory(A32EmitContext& ctx, IR::Inst* inst) {
...
+    Xbyak::Label end;
+
+    const auto dest_ptr = r13 + vaddr;
+
+    code.mov(tmp, Common::BitCast<u64>(GetExclusiveMonitorAddressPointer(conf.global_monitor, conf.processor_id)));
+
+    code.mov(status, u32(1));
+    code.cmp(code.byte[r15 + offsetof(A32JitState, exclusive_state)], u8(0));
+    code.je(end);
+    code.cmp(qword[tmp], vaddr);
```
FernandoS27, line 498: race condition, use separate storage.

src/dynarmic/backend/x64/a32_emit_x64_memory.cpp
merryhime: intentional behaviour

src/dynarmic/backend/x64/a32_emit_x64_memory.cpp
merryhime: using separate storage would result in incorrect behaviour

PR #665, opened 2022-03-13
build: remove extra include path for system vixl

Tachi107:
As far as I know the only pkg-config file provided by Vixl is the one generated by Meson when applying my yet to be merged patch.

That extra include path was needed because I mistakenly thought that adding `vixl` as an include subdirectory was not necessary, but I fixed it in my latest revision - more details here: https://github.com/Linaro/vixl/pull/7#discussion_r778167004

The fix already landed in Debian and Ubuntu, that as far as I know are the only Linux distros that ship my patch, so manually adding that include directory shouldn't be necessary anymore


merryhime:
cheers.

PR #666, opened 2022-03-20
backend/x64: Actually enforce memory ordering

merryhime:
Here we (over?) enforce memory ordering for memory accesses of type `AccType::ORDERED` or stronger.

If we could assume a well-behaved guest, we would only need to have a store-load barrier on either before the acquire or after the release. We however can't really make the assumption of well-behaved code, thus we must emit a barrier on both.

We however do make the assumption that we will never have a need to have a barrier against non-temporal loads. This assumption is sound because we do not emit any non-temporal load instructions and do not have any intention of doing so. For similar reasons, we do not support the use of this library in devices operating on time machines.

The memory barriers are implemented using `lock xadd 0` for loads, `xchg` (with implicit `lock` prefix) for stores, and if we can't use those instructions for whatever reason, with an appropriately placed `mfence`.

There are a few more things to do before this is complete:

- [x] Fix tests
- [x] General code cleanup
- [x] Implement in A32 frontend
- [x] Ensure 128-bit loads are atomic (`movaps` is actually only guaranteed to be atomic on Intel processors supporting AVX; however also see https://rigtorp.se/isatomic/)
- [x] Remove redundant `mfence`s: An `mfence` when `lock cmpxchg` is used as that already implies a bidirectional barrier.
- [x] Consider whether A32 and A64 memory code in backend/x64 should be merged for ease of mantanence.


merryhime:
Most of the changes in this PR are now on master. One last TODO is to ensure 128-bit loads and stores are atomic on certain CPUs.

merryhime:
All tasks complete on master.

PR #668, opened 2022-03-21
cmake: Fix unicorn and llvm

abouvier:
* don't require `Unicorn` if tests are disabled
* fix variable expansion when `DYNARMIC_USE_LLVM` is enabled


merryhime:
thanks - is it possible to show why those quotation marks are required for my own learning?

lioncash (approved):

PR #670, opened 2022-03-27
backend/x64: Use upper EVEX registers as scratch space

Wunkolo:
[AVX512 adds an additional **16** simd registers](https://en.wikipedia.org/wiki/EVEX_prefix), for a total of 32 simd
registers, accessible by utilizing EVEX encoded instructions. Rather
than using the `ScratchXmm` function, adding additional
register-pressure and spilling, AVX512-enabled contexts can just
directly use `xmm{16-31}` registers as intermediate scratch registers.


src/dynarmic/backend/x64/emit_x64_vector.cpp
```diff
@@ -628,21 +631,23 @@ void EmitX64::EmitVectorArithmeticVShift64(EmitContext& ctx, IR::Inst* inst) {
...
 
-        code.vmovdqa(tmp, code.MConst(xword, 0x00000000000000FF, 0x00000000000000FF));
-        code.vpxor(right_shift, right_shift, right_shift);
+        code.vmovdqa32(tmp, code.MConst(xword, 0x00000000000000FF, 0x00000000000000FF));
+        code.vpxorq(right_shift, right_shift, right_shift);
         code.vpsubq(right_shift, right_shift, left_shift);
 
         code.vpsllq(xmm0, left_shift, 56);
+        const Xbyak::Opmask mask = k1;
+        code.vpmovd2m(mask, xmm0);
```
merryhime, line 643: probably `vpmovq2m`

merryhime:
I feel this is somewhat of a stopgap for use of the full AVX512 register file (not that we really need to do so rn…)

Wunkolo:
It definitely is. I have tried to get it to use the full register-space before on a local branch up to the point that where we would have to conditionally emit the EVEX-version of every simd instruction.

PR #676, opened 2022-04-23
cmake: build static externals as PIC

abouvier:
Maybe a better fix for https://github.com/merryhime/dynarmic/issues/675, you decide :p
This will work with future externals too.


merryhime:
Ah, good point, didn't think of this.

PR #684, opened 2022-06-16
Waitpkg

### LINKED ISSUE #672: x64: `WAITPKG`-based spinlocks
The `waitpkg` cpuid flag indicates the availability of some user-space instructions for monitoring an address region and waiting for a specific memory-region to be written to, in particular `umonitor` and `umwait`. These instructions are found in tremont, sapphire rapids, and alderlake.
https://www.felixcloutier.com/x86/umonitor
https://www.felixcloutier.com/x86/umwait

I proposed the additions for xbyak here: https://github.com/herumi/xbyak/issues/143

These instructions can possibly be used to [accelerate spinlocks](http://liujunming.top/2022/03/01/Introduction-to-umonitor-umwait-tpause-instructions/).

Wunkolo:
This implements an [optimal method of spin-locking](http://liujunming.top/2022/03/01/Introduction-to-umonitor-umwait-tpause-instructions/) that utilizes the `waitpkg` extension(available since `tremont`, `sapphire rapids`, and `alder lake`) to designate the time and powerstate that the processor is allowed to transition into when spinning and waiting for the lock to be written to.

Since this is primarily used for exclusive memory operations, I've allowed it to use power-state `C0.1`.

image

Resolves #672


merryhime:
`umonitor` likely needs to be within the spinloop

src/dynarmic/common/spin_lock_x64.cpp
```diff
@@ -48,7 +72,7 @@ SpinLockImpl::SpinLockImpl() {
 
     code.align();
     lock = code.getCurr<void (*)(volatile int*)>();
-    EmitSpinLockLock(code, ABI_PARAM1, code.eax);
+    EmitSpinLockLock(code, ABI_PARAM1, code.eax, false);
```
merryhime, line 75: tbh I don't see why this can't feature detect

Wunkolo: The `code` variable here is an `Xbyak::CodeGenerator` rather than the usual `BlockOfCode` so `HasHostFeature` is not available here.

merryhime:
reminder to self: use of `eax:edx` in context where they may not be avaliable for use

Wunkolo:
Haven't been able to test this stuff myself due to my lack of hardware at the moment. But still a valuable addition to make once the implementation can be verified. Also the usage of `edx:eax` here _might_ be unsafe. Since it does not interface with the register-allocator.

PR #685, opened 2022-06-18
emit_x64_{vector_}floating_point: Add AVX512 implementation for `DenormalsAreZero`

Wunkolo:
Both single and double precision floating point numbers as well as the
packed and unpacked version of this instruction will be able to use the
same memory constant. This takes advantage of the fact that `VFIXUPIMM*`
doesn't just copy from the source, [but it will convert to `0.0` if it
turns out that it is a denormal and the `MXCSR.DAZ` flag is set](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#techs=AVX_512&text=vfixupimmps&ig_expand=3119,3119).

```
tsrc[31:0]←((src1[30:23] = 0) AND (MXCSR.DAZ =1)) ? 0.0 : src1[31:0]
...
CASE(token_response[3:0]) {
    ...
    0001: dest[31:0]←tsrc[31:0]; ; pass through src1 normal input value, denormal as zero
    ...

### CODE REVIEW

File: src/dynarmic/backend/x64/emit_x64_floating_point.cpp [hunk 1]
```diff
@@ -79,6 +80,24 @@ constexpr u64 f64_max_s64_lim = 0x43e0000000000000u;  // 2^63 as a double (actua
...
+                FpFixup::Norm_Src,
+                FpFixup::Norm_Src,
+                FpFixup::Norm_Src,
+                FpFixup::Norm_Src,
+                FpFixup::Norm_Src,
+                FpFixup::Norm_Src,
+                FpFixup::Norm_Src,
+                FpFixup::Norm_Src);
+            const Xbyak::Xmm tmp = xmm16;
+            code.movaps(tmp, code.MConst(xword, u64(denormal_to_zero), u64(denormal_to_zero)));
```
merryhime, line 94: Using `xmm16` with `movaps` is not supported.

Wunkolo: You're right. Turns out this was also not handling _both_ single and double precision floats correctly as the 32-bit `fixup`-value should be replicated across each of the `fsize`-large lanes.

PR #689, opened 2022-06-28
constant_pool: Use `std::span` to manage pool

Wunkolo:
Simplifies some raw pointer arithmetic and type-usage into the new
`ConstantT` type.


merryhime:
I don't think the copyright notice requries updating. Afaik, the year on a copyright notice should be the year of first publication.

src/dynarmic/backend/x64/constant_pool.cpp
```diff
@@ -14,22 +14,22 @@
 namespace Dynarmic::Backend::X64 {
 
 ConstantPool::ConstantPool(BlockOfCode& code, size_t size)
-        : code(code), pool_size(size) {
+        : code(code), pool_end(0) {
```
merryhime, line 17: s/`pool_end`/`insertion_point`/

merryhime (commented):
Been thinking about this PR.

Main reason I used `std::memcpy` here with `u8*` pointers was because of thoughts regarding strict-aliasing, though I'm fairly certain that doesn't apply here and I'm being overly paranoid.

The `void*` pointers are only used in JIT-ted code.

@lioncash Any thoughts?

PR #690, opened 2022-06-28
block_of_code: Refactor `MConst` to `Xmm{B}Const`

Wunkolo:
`MConst` is refactored into `XmmConst` to clearly communicate the
addressable space of the newly allocated 16-byte memory constant.
`GetVectorOf` is elevated into a globally available `XmmBConst` function
that "broadcasts" bits of the input-value into n-bit elements that span
the width of the Xmm-constant.

`emit_x64_floating_point` will utilize the same 16-byte
broadcasted-constants to encourage more cache-hits within the
constant-pool between vector and non-vector code.


merryhime (approved):
I'm mildly surprised things don't break on testing actually with the broadcasted constants.

Historically, emitted code may have assumed upper parts of registers were zero. i.e. if there was a 64-bit value in an xmm register, some vector code might rely on the assumption that the upper 64-bits were zero.

I think I stamped most of that out in most code involving GPRs, but haven't really audited the code for this.

PR #697, opened 2022-07-14
ARM64 Backend

merryhime:
Framework of ARM64 backend. Aim is for a working A32 support to be done first.

This also involves a redesign of how backends are done, to potentially support multiple threads sharing emulated address spaces, which I plan to backport to the x64 backend.

Very incomplete.


src/dynarmic/backend/arm64/emit_arm64.h
```diff
@@ -21,12 +21,14 @@ class Block;
 
 namespace Dynarmic::Backend::Arm64 {
 
+using CodePtr = char*;
```
Wunkolo, line 24: I'm an [std::byte](https://en.cppreference.com/w/cpp/types/byte) and [std::span](https://en.cppreference.com/w/cpp/container/span) enjoyer. Maybe this can be a moment to take advantage of these while these new types are being introduced.

merryhime:
Reminder: Handle fpcr_controlled in FPVector.

merryhime:
Things to do:

- [ ] Properly implement terminals
- [ ] Properly implement ranged cache invalidation

PR #703, opened 2022-08-17
backend/arm64: implement A64SetCheckBit based on A32SetCheckBit

FreddyFunk:
This is my first attempt on implementing a tiny function of the arm64 backend for A64.

Since the A32/x64 and A64/x64 both have an identical implementation I guessed that the arm64 backend might behave identically. In case this assumption is true, all credit for the source code goes to @MerryMage. In this assumption is false, I'd love to get some guidance to get this PR right.


merryhime:
IIRC the reason why they're separate instructions is a historical artifact from when check_bit was a member of A32JitState/A64JitState.

merryhime (approved):
Will merge when appropriate.

PR #704, opened 2022-08-17
tests: remove unused default iterations in TestThumb

FreddyFunk:


FreddyFunk:
Not sure why the macOS test case failed. Could it be a false positive and a test re-run could fix this?

PR #705, opened 2022-08-18
backend/arm64: multi-thread safe A32AddressSpace

FreddyFunk:
This PR aims to make A32AddressSpace multi-thread safe from an invalidation and code emission standpoint.

Not sure if the following scenario is possible. If so, then we might need to consider a more global mutex variable.

1. Thread A: Emits a new code block "X" (thread safe)
2. Thread B: Waits for "step 1." to be completed due to the unique_lock
3. Thread B: Invalidates new code block "X" (thread safe) due to a cache invalidation
4. Thread A: Tries to run new code block "X" on host, but the block is not in memory anymore.

### DIFF
--- a/src/dynarmic/backend/arm64/a32_address_space.cpp
+++ b/src/dynarmic/backend/arm64/a32_address_space.cpp
@@ -122,6 +122,8 @@ IR::Block A32AddressSpace::GenerateIR(IR::LocationDescriptor descriptor) const {
 }
 
 CodePtr A32AddressSpace::Get(IR::LocationDescriptor descriptor) {
+    std::shared_lock lock{address_space_lock};
+
     if (const auto iter = block_entries.find(descriptor.Value()); iter != block_entries.end()) {
         return iter->second;
     }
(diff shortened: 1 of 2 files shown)


merryhime:
Would probably need to be more complex than this.

- Invalidation at a basic block level has not been implemented and would need to be considered.
- Invalidation of individual basic blocks would need to consider unlinking/relinking each basic block.
- Invalidation of would need to consider the possibility of other threads currently executing code
  - The way this would probably be handled performantly is by triggering a cache invalidation interrupt on all cores; ensuring all cores are idled before performing an invalidation.
  - Would also need to consider the possibility of threads waiting on the compilation lock, and ensure this doesn't result in a deadlock.
- For performance, `Get` must not take a lock. This would likely be handled with a concurrent spmd hashmap.
  - Half the reason I implemented [my own hashmap](https://github.com/merryhime/mcl/blob/main/include/mcl/container/hmap.hpp) was so I could then write a concurrent one with the properities I want, and also inline an implementation to the dispatcher without needing to call back into C++.

FreddyFunk:
Thanks a lot for the guidance! :) I'll focus on the invalidation part first.

PR #718, opened 2022-11-14
backend/arm64: Add MSVC C++ ABI devirtualization

Wunkolo:
Ported over from backend/x64

Depends on https://github.com/merryhime/oaknut/pull/1

This gets Dynarmic building and running on Windows on Arm.

Passes all A32 unit tests on my ThinkPad x13s:
```
0.000 s: FPToFixed
0.000 s: ResidualErrorOnRightShift
2.058 s: ResidualErrorOnRightShift Randomized
0.001 s: FPUnpack Tests
0.000 s: FPRound Tests
3.172 s: FPUnpack<->FPRound Round-trip Tests
0.000 s: FPRound (near zero, round to posinf)
0.002 s: Disassemble branch instructions
0.004 s: Disassemble data processing instructions
0.000 s: Disassemble half-word multiply and multiply accumulate instructions
0.000 s: Disassemble multiply and multiply accumulate instructions
0.000 s: Disassemble synchronization primitive instructions
0.000 s: Disassemble load / store multiple instructions
0.108 s: arm: Opt Failure: Const folding in MostSignificantWord
0.105 s: arm: Unintended modification in SetCFlag
0.103 s: arm: shsax (Edge-case)
0.104 s: arm: uasx (Edge-case)
0.103 s: arm: smuad (Edge-case)
0.102 s: arm: Test InvalidateCacheRange
0.101 s: arm: Step blx
0.101 s: arm: Step bx
0.103 s: arm: Test stepping
0.104 s: arm: Test stepping 2
0.103 s: arm: Test stepping 3
0.105 s: arm: PackedAbsDiffSumS8
0.103 s: arm: vclt.f32 with zero
0.103 s: arm: vcvt.s16.f64
0.102 s: arm: Memory access (fastmem)
0.103 s: arm: vmsr, vcmp, vmrs
0.102 s: arm: Test coprocessor (Read TPIDRURO)
0.107 s: arm: Test coprocessor (Read TPIDRURW)
0.108 s: arm: Test coprocessor (Write TPIDRURW)
0.103 s: arm: Test coprocessor (DMB)
0.102 s: thumb: lsls r0, r1, #2
0.102 s: thumb: lsls r0, r1, #31
0.103 s: thumb: revsh r4, r3
0.104 s: thumb: ldr r3, [r3, #28]
0.103 s: thumb: blx +#67712
0.102 s: thumb: bl +#234584
0.106 s: thumb: bl -#42
0.113 s: thumb: Opt Failure: Get/Set Elimination for Flags
0.107 s: thumb: Opt Failure: Get/Set Elimination for Flags 2
===============================================================================
All tests passed (200536 assertions in 42 test cases)
```


merryhime:
Some nitpicking:

1. Given the [ARM C++ ABI](https://github.com/ARM-software/abi-aa/blob/main/cppabi64/cppabi64.rst#representation-of-pointer-to-member-function) specifies the behaviour I feel we can assume this is the default behaviour and to treat windows as an outlier:

```cpp
#if defined(_WIN32) && !defined(__MINGW64__)
    return DevirtualizeWindows<mfp>(this_);
#else
    return DevirtualizeStandard<mfp>(this_);
#endif
```

(Note I've not verified if `mingw-w64-clang-aarch64-toolchain` ignores the Windows C++ ABI.)

2. Technically it's not Itanium, so naming it such seems unwise. Given it's a published ARM standard and should be the platform default, naming it `Default` or `Standard` seems more suitable.

Wunkolo:
Was able to verify that pretty much any windows-facing clang will implement the Windows C++ ABI by default, including the member-pointer caveat.

Looks like clang's MSVC ABI compatibility is documented here.
https://clang.llvm.org/docs/MSVCCompatibility.html

> Member pointers: Mostly complete. Standard C++ member pointers are fully implemented and should be ABI compatible. Both [#pragma pointers_to_members](https://msdn.microsoft.com/en-us/library/83cch5a6.aspx) and the [/vm](https://msdn.microsoft.com/en-us/library/yad46a6z.aspx) flags are supported. However, MSVC supports an extension to allow creating a [pointer to a member of a virtual base class](https://llvm.org/PR15713). Clang does not yet support this.

src/dynarmic/backend/arm64/devirtualize.h
```diff
@@ -5,6 +5,8 @@
 
 #pragma once
 
+#include <utility>
```
merryhime, line 8: Unnecessary include? Otherwise LGTM.

Wunkolo: Oops. Residue from `ThunkBuilder`.

PR #724, opened 2022-11-22
CMake: Add option to use precompiled headers

ameerj:
This adds a CMake option to make use of precompiled headers. 

Currently, the only 2 headers that are precompiled are `ir/ir_emitter.h` and `<xbyak/xbyak.h>` as my analysis showed these are the two commonly used headers which lead to the longest build times.

On my machine build times were halved building dynarmic standalone. Projects which depend on dynarmic may see a noticeable build time improvement.


merryhime:
> [35/258] Building C object src/dynarmic/CMakeFiles/dynarmic.dir/cmake_pch.h.pch

The precompiled header is being compiled with a C compiler.
Perhaps consider specifying the headers as `"$<$<COMPILE_LANGUAGE:CXX>:${CMAKE_CURRENT_SOURCE_DIR}/ir/ir_emitter.h>"` and `"$<$<COMPILE_LANGUAGE:CXX>:<xbyak/xbyak.h$<ANGLE-R>>"` instead.

ameerj:
Thanks for catching that it was compiled as a C header. CI is happy now.

PR #725, opened 2022-11-24
cmake: multiple simplifications

abouvier:
- Always use imported target for external libraries, including `Boost::boost` created by [FindBoost.cmake](https://github.com/Kitware/CMake/blob/93ded7304aea9848352ce843d98fbf64023be11f/Modules/FindBoost.cmake#L132-L133).
- Remove the `DYNARMIC_NO_BUNDLED_*` variables and prefer system libraries by default, with fallback on externals when not found.
- Using `$<BUILD_INTERFACE:(...)>` was an ugly hack to hide the fact that subprojects lacked install/export rules. The install/export rules of a subproject should not be guarded behind a `MASTER_PROJECT` variable, but disabled with the `EXCLUDE_FROM_ALL` option of `add_subdirectory` when including the project.
- Building fails on `arm64` because `oaknut` lacks install/export rules. Every project should have unconditional install/export rules available (`oaknut`, `mcl`, `dynarmic` and even `fmt`!).
- Link `dynarmic` with `merry::mcl` privately because public headers in `dynarmic/interface` don't include `mcl` headers.


externals/CMakeLists.txt
```diff
@@ -10,56 +10,57 @@ endif()
 
 # catch
 
-if (DYNARMIC_TESTS AND NOT TARGET catch)
-    add_library(catch INTERFACE)
-    target_include_directories(catch INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/catch/include>)
+if (NOT TARGET Catch2::Catch2)
+    if (DYNARMIC_TESTS)
```
Tachi107, line 14: This can probably be simplified as:

```suggestion
if (NOT TARGET Catch2::Catch2 AND DYNARMIC_TESTS)
```

Tachi107:
>  * Remove the `DYNARMIC_NO_BUNDLED_*` variables and prefer system libraries by default, with fallback on externals when not found.

I think that it'd be nice to mention somewhere that externals can be preferred by using [`-DCMAKE_DISABLE_FIND_PACKAGE_PackageName=true`](https://cmake.org/cmake/help/latest/variable/CMAKE_DISABLE_FIND_PACKAGE_PackageName.html)

>  * Using `$<BUILD_INTERFACE:(...)>` was a ugly hack to hide the fact that subprojects lacked install/export rules. The install/export rules of a subproject should not be guarded behind a `MASTER_PROJECT` variable, but disabled with the `EXCLUDE_FROM_ALL` option of `add_subdirectory` when including the project.

Yeah, this was kinda [surprising to me](https://github.com/pistacheio/pistache/pull/1073) too. Still, note that guarding installation behind [`PROJECT_IS_TOP_LEVEL`](https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html) (or `MASTER_PROJECT`) [does not create issues](https://github.com/pistacheio/pistache/pull/1091) to my understanding.

>  * Building fails on `arm64` because `oaknut` lacks install/export rules. Every project should have unconditional install/export rules available (`oaknut`, `mcl`, `dynarmic` and even `fmt`!).

Could you expand a bit on this? Building on arm64 [works fine on Debian](https://buildd.debian.org/status/logs.php?pkg=dynarmic&arch=arm64).

externals/CMakeLists.txt
abouvier: Yes, but I was trying to mimic the existing condition for `xbyak` and `Zydis`!

abouvier:
> I think that it'd be nice to mention somewhere that externals can be preferred by using [`-DCMAKE_DISABLE_FIND_PACKAGE_PackageName=true`](https://cmake.org/cmake/help/latest/variable/CMAKE_DISABLE_FIND_PACKAGE_PackageName.html)

OK, I added a comment above the first `find_package`.

> Yeah, this was kinda [surprising to me](https://github.com/pistacheio/pistache/pull/1073) too. Still, note that guarding installation behind [`PROJECT_IS_TOP_LEVEL`](https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html) (or `MASTER_PROJECT`) [does not create issues](https://github.com/pistacheio/pistache/pull/1091) to my understanding.

Guarding the install rules of a subproject prevents the top project from using the imported targets created by the subproject. That's why I need to set `FMT_INSTALL` and `MCL_INSTALL` to `TRUE`.

> Could you expand a bit on this? Building on arm64 [works fine on Debian](https://buildd.debian.org/status/logs.php?pkg=dynarmic&arch=arm64).

Yes it works with the current code, because `oaknut` is linked with the hack `$<BUILD_INTERFACE:merry::oaknut>`.

merryhime:
Thanks! Someday one may learn to CMake properly.

bunnei:
@merryhime this seems to break on CMake 3.22.1, which is the latest version available with Android Studio

```
CMake Error: install(EXPORT "dynarmicTargets" ...) includes target "dynarmic" which requires target "fmt" that is not in any export set.
CMake Error: install(EXPORT "dynarmicTargets" ...) includes target "dynarmic" which requires target "mcl" that is not in any export set.
```

abouvier:
That means the install rules of `fmt` and `mcl` are not executed in your build. Probably because the variables `FMT_INSTALL` and `MCL_INSTALL` are not set. Maybe because of [this](https://cmake.org/cmake/help/latest/policy/CMP0077.html)?

Tachi107:
I still don't fully understand why install rules must me always available. Imported targets should only be used when the project is _imported_ from somewhere else, not when the project is built as a subproject, right?

abouvier:
`find_package` creates imported target like `fmt::fmt`, but `add_subdirectory` only create internal library like `fmt`.  If it's possible to use either system libraries or externals, you need to have a common name to link with. You can either:

1. alias the imported target `fmt::fmt` to a fake internal library `fmt` after `find_package` and use internal names everywhere
2. alias the internal `fmt` to a fake imported target `fmt::fmt` after `add_subdirectory` (some project even [do this in their root CMakeLists.txt](https://github.com/merryhime/dynarmic/blob/2d3c8c5724800534f68e801e3f18b13a4d400e22/externals/robin-map/CMakeLists.txt#L9) so install rules are not needed) and use imported names everywhere
3. have install rules in the subproject to make the imported target available in the build tree and use imported names everywhere

Tachi107:
Aliasing the library obtained via `add_subdirectory()` to the "imported name" looks to me like the simplest and less troublesome approach. As opposed to when doing what you mention in the point 1, CMake checks that targets containing `::` in their name correspond to an actual imported target or aliased library, instead of silently passing `-lname` to the linker - in a sense, CMake acknowledges this practice and accounts for its usage. (see CMP0028: https://cmake.org/cmake/help/latest/policy/CMP0028.html)

merryhime:
> @merryhime this seems to break on CMake 3.22.1, which is the latest version available with Android Studio
> 
> ```
> CMake Error: install(EXPORT "dynarmicTargets" ...) includes target "dynarmic" which requires target "fmt" that is not in any export set.
> CMake Error: install(EXPORT "dynarmicTargets" ...) includes target "dynarmic" which requires target "mcl" that is not in any export set.
> ```

If you are not building from a clean cache, remove `FMT_INSTALL` and `MCL_INSTALL` from your existing CMake cache.

PR #732, opened 2022-12-27
ci: Add Android aarch64 build

Macdu:
New commits often break Dynarmic when building it for Android (clang NDK compiler + Android headers).
As such, I think adding an Android (aarch64) build to the ci would be a good idea to prevent this from happening.

The Android build is not tested (only the linux gcc one is), but making sure it is able to build itself should already be enough from my point of view.

### DIFF
--- a/.github/workflows/aarch64.yml
+++ b/.github/workflows/aarch64.yml
@@ -9,15 +9,14 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [ ubuntu-latest ]
+        platform: [ linux, android ]
       fail-fast: false
 
-    runs-on: ${{matrix.os}}
+    runs-on: ubuntu-latest
 
     steps:
 
       - name: Install build dependencies
-        if: ${{matrix.os == 'ubuntu-latest'}}
         run: >
           sudo apt-get update &&
           sudo apt-get install
@@ -30,6 +29,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Ccache
+        if: matrix.platform == 'linux'
         uses: hendrikmuhs/ccache-action@v1.2
 
       - name: Checkout ext-boost repo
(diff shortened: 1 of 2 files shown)


tests/CMakeLists.txt
```diff
@@ -101,6 +101,9 @@ endif()
 
 create_target_directory_groups(dynarmic_tests)
 
+if(CMAKE_SYSTEM_NAME STREQUAL "Android")
+    target_link_libraries(dynarmic_tests PRIVATE log)
```
merryhime, line 105: Why is log required?

merryhime (commented):
Looks fine to be otherwise

tests/CMakeLists.txt
Macdu: Catch2 has builtin support for android:
[https://github.com/merryhime/dynarmic/blob/1fa1935d7c015767cf211c441578b25e3517c43f/externals/catch/include/catch2/catch.hpp#L10374](https://github.com/merryhime/dynarmic/blob/1fa1935d7c015767cf211c441578b25e3517c43f/externals/catch/include/catch2/catch.hpp#L10374)
And uses the android function `__android_log_write`. For this symbol to be defined and the linker not to fail (see [https://github.com/Macdu/dynarmic/actions/runs/3783711085/jobs/6432474542](https://github.com/Macdu/dynarmic/actions/runs/3783711085/jobs/6432474542)), the android log library needs to be linked.

merryhime: Should be added to externals/CMakeLists.txt in that case: 
```
if(CMAKE_SYSTEM_NAME STREQUAL "Android")
    target_link_libraries(Catch2::Catch2 INTERFACE log)
endif()
```

Macdu: done

PR #738, opened 2022-12-31
cmake: allow shared llvm linking

abouvier:
`LLVM` is now linked to the `dynarmic` target instead of the whole project. And if `dynarmic` is compiled as a shared library, then `LLVM` is linked dynamically.

The condition in `$<$<BOOL:DYNARMIC_USE_LLVM>:${llvm_libs}>` was always true by the way, but fortunately `$llvm_libs` was empty when `DYNARMIC_USE_LLVM` was disabled. It should have been ``$<$<BOOL:${DYNARMIC_USE_LLVM}>:${llvm_libs}>``.


merryhime:
Thanks for the heads up.

PR #748, opened 2023-02-17
backend/x64: Utilize extended EVEX registers as Spill-Space

Wunkolo:
AVX512 adds an additional 16 `{x,y,z}mm` registers to the
register-state. We only address the `xmm` part of these registers, so
this allows an additional 256-bytes of fast register-memory. These
registers cannot be utilized immediately without updating all of our
instructions to conditionally utilize EVEX encoding to be able to
address these additional registers. Instead, these additional registers
can be utilized as fast additional spill-space before having to utilize
system memory.


merryhime:
calls will need these dumped to memory spill

Wunkolo:
Intel APX doubles the general purpose registers similarly to how AVX512 doubles the SIMD registers. So this might be the premise for a similar register-extension pattern for APX.
https://www.intel.com/content/www/us/en/developer/articles/technical/advanced-performance-extensions-apx.html

PR #753, opened 2023-04-21
Fix dynarmic on iOS

Steveice10:
Defers initialization of ```SpinLock``` code block until used. This is because the prevailing JIT workarounds for iOS usually set up JIT capabilities on the process just after launch, meaning when we launch we don't yet have JIT access. Deferring until dynarmic is actually used allows us to get that permission first before emitting the ```SpinLock``` code.


externals/oaknut/include/oaknut/code_block.hpp
```diff
@@ -64,14 +69,22 @@ class CodeBlock {
     void protect()
     {
 #if defined(__APPLE__)
+    #if TARGET_OS_IPHONE
+        mprotect(m_memory, m_size, PROT_READ | PROT_EXEC);
```
SachinVin, line 73: Does iOS enforce W^X?

Steveice10: To my knowledge, yes.

Steveice10:
Rebased and hopefully fixed Windows build issue. Checked MSVC docs and it should support [[unlikely]] so I've switched over to using that.

Steveice10:
There appears to be an occasional test crash going back a while in the master branch actions history, so it’s unclear to me if the single failure here is related to this PR or not.

PR #760, opened 2023-08-20
CMakeLists: disable -Warray-bounds for GCC

comex:
The reason why is explained in the comment.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,6 +104,15 @@ else()
              -Wfatal-errors)
     endif()
 
+    if (CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+        # GCC produces bogus -Warray-bounds warnings from xbyak headers for code paths that are not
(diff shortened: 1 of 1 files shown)


abouvier:
Isn't it [already fixed](https://github.com/herumi/xbyak/commit/9c8fb81db8d71bb5a1490801bdcc72479b8c52cb) directly in `xbyak`?

merryhime:
Hey comex.

Related issues:
* herumi/xbyak#137
* https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104603

Would prefer this to be fixed upstream where possible, alternatively for warning disable coverage to be minimized.

Alternatively, does adding the `SYSTEM` argument  to `add_subdirectory(xbyak)` correct this downstream? (Which should ignore all warnings in related headers.)

merryhime:
Hopefully until GCC fixes this bogus warning.

PR #762, opened 2023-10-07
Support NetBSD and OpenBSD

scribam:
Commit https://github.com/fmtlib/fmt/commit/3c5464ba1cbac7fa84009a5ec207977208be69e2 is required to build on OpenBSD, that is why I have updated `fmt`


merryhime:
Many thanks, appreciated.

PR #763, opened 2023-11-15
A32: Introduce PreCodeReadHook

kynex7510:
This commit adds a new translator callback, which allows the user to intercept the translation of an instruction, replacing it with custom IR code.


src/dynarmic/interface/A32/config.h
```diff
@@ -65,6 +65,12 @@ struct UserCallbacks : public TranslateCallbacks {
     // Memory must be interpreted as little endian.
     std::optional<std::uint32_t> MemoryReadCode(VAddr vaddr) override { return MemoryRead32(vaddr); }
 
+    // This function is called before the instruction at pc is read.
+    // IR code can be emitted by the callee prior to instruction handling.
+    // By returning true the callee precludes the translation of the instruction;
```
merryhime, line 70: Inverting the meaning of return value may be preferable. (Return true/something positive to continue.)

FernandoS27:
Out of curiosity, coukd you share an special case where a user could use this?

kynex7510:
> Out of curiosity, coukd you share an special case where a user could use this? 

In my case, I want to handle ELF import relocation by generating a custom stub, with the actual address pointing to invalid (unmapped) memory. `PreCodeTranslationHook` wouldn't work for this as the instruction would be read first, causing an exception.

PR #764, opened 2023-11-18
CMakeLists: Fix build with /WX and clang-cl on non-VS generators.

zmt00:
Testing if the C++ compiler is clang-cl via `CMAKE_VS_PLATFORM_TOOLSET` seems to only detect builds done with Visual Studio generators. As such, building dynarmic on Windows with
```
cmake .. -GNinja -DCMAKE_TOOLCHAIN_FILE="C:\dev\vcpkg\scripts\buildsystems\vcpkg.cmake" -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl
```
failed for me, as several command-line options unimplemented by clang-cl caused the build to fail. In contrast, building with the Visual Studio generator via
```
cmake .. -G"Visual Studio 17 2022" -DCMAKE_TOOLCHAIN_FILE="C:\dev\vcpkg\scripts\buildsystems\vcpkg.cmake" -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl
```
worked. (The toolchain file is my Boost installation.)

Detecting clang-cl with this method works with both VS generators and ninja, and I presume will work with any other non-VS generators.


merryhime:
Seems reasonable.

PR #765, opened 2023-12-02
tests/A64, emit_x64_vector: Add unit tests, SSE4.1 implementation of VUZP{1,2}.S

zmt00:
`insertps` has a latency and throughput similar to `unpcklps` on every platform that I have checked, but does not require any followup instructions. 

I don't believe there's any further optimizations available for UZP{1,2}.4S, so I added tests for them as well, just to be complete.


zmt00:
Accidentally made a (bad) typo at the last second before pushing, in the no-cpu-optimizations fallback branch - pushed a fix now. (Very thankful I added some tests.)

PR #766, opened 2023-12-10
tests/A64, emit_x64_vector: Add unit tests, optimize, VUZP{1,2}.H

zmt00:
For the SSE4.1 `VectorDeinterleaveEven16` implementation, `pblendw` has a latency of 1 and a throughput of 1 or better, similarly to `pslld` and `psrad`, with fewer instructions. The latency of `packusdw` should be the same as `packssdw` on ~all CPUs.

The new `VectorDeinterleaveLower{Even,Odd}16` implementations should be clear wins,


zmt00:
Force-pushed to fix some clang-tidy formatting errors.

merryhime (approved):

PR #772, opened 2023-12-31
backend/rv64: Add initial RISC-V framework

ksco:
👋 Hi, I would like to start adding a RISC-V backend for dynarmic. The plan is to first implement the A32 backend with host architecture RV64GC. Will dynarmic accept such a new backend?

As for this PR, the RISC-V backend is now compilable, but nothing works.


merryhime:
An rv64 backend would be interesting. I would happily merge it piecewise in a separate branch until usable.
Out of interest, is there any particular physical target in mind?

ksco:
> I would happily merge it piecewise in a separate branch until usable.

That's great! A separate branch is reasonable.

> Out of interest, is there any particular physical target in mind?

I own a LicheePi 4A dev board and a Milk-V Pioneer Box, but their performance is nothing compared to common x86-64/AArch64 hardware.

PR #775, opened 2024-01-02
block_of_code: Refactor `MConst` to `{,B}Const`

Wunkolo:
This is a redo of https://github.com/merryhime/dynarmic/pull/690 with a
much smaller foot-print to introduce a new pattern while avoiding the
initial bugs(https://github.com/merryhime/dynarmic/commit/5d9b720189a64eec7f35f844320d0b30ca3997f3).

**B**roadcasts a value as an **Xmm**-sized **Const**ant. Intended to
eventually encourage more cache-hits within the constant-pool between vector
and non-vector code.


src/dynarmic/backend/x64/block_of_code.h
```diff
@@ -22,6 +22,7 @@
 #include "dynarmic/common/cast_util.h"
 #include "dynarmic/interface/halt_reason.h"
 #include "dynarmic/ir/cond.h"
+#include "mcl/bit/bit_field.hpp"
```
lioncash, line 25: This should be up by `<mcl/stdint.h>` and also be using `<>`, since it's an external header

merryhime:
This is nitpicking but I feel the name should probably not be `XmmConst`. The reason it was named `MConst` (which probably derives from "memory constant") was because this is equivalent to a function of the same name in dolphin.

This function can be used with memory constants of sizes other than `xmm`, such as something of size `dword` or `qword`.

merryhime (approved):
Will run tests later prior to merging.

PR #777, opened 2024-01-07
backend/rv64: Minor changes before implement register allocator

ksco:
I'm following how the AArch64 backend is implemented, next thing to implement would be the register allocator.


src/dynarmic/backend/riscv64/a32_address_space.h
```diff
@@ -36,6 +36,20 @@ class A32AddressSpace final {
 
     void EmitPrelude();
 
+    template<typename T>
+    T GetMemPtr() const {
+        static_assert(std::is_pointer_v<T> || std::is_same_v<T, uptr> || std::is_same_v<T, sptr>);
+        return reinterpret_cast<T>(const_cast<u8*>(as.GetBufferPointer(0)));
```
lioncash, line 42: I would extremely prefer if the interface were adapted to handle const and non-const retrieval if possible. If an API improvement in the future requires breaking this code, I will do it, because the biscuit API clearly lays out both a non-const and const getter to satisfy concerns

ksco: Thank you for the review, force-pushed. I suppose this is what you suggested? (I'm very bad at modern C++, please don't hesitate to point out any unusual practices of mine.)

lioncash: it's all good. Ideally if we were in a scenario where deduced [this](https://devblogs.microsoft.com/cppblog/cpp23-deducing-this) was a thing, it'd be exceptionally good, but dynarmic really isn't a state to require c++23 atm

ksco: Ah yes, the "deduced this" thing suits well for this use case indeed.

src/dynarmic/backend/riscv64/a32_address_space.cpp
```diff
@@ -60,43 +62,49 @@ void* A32AddressSpace::GetOrEmit(IR::LocationDescriptor descriptor) {
...
         as.FLD(FPR{i}, 32 + i * 8, sp);
     }
     as.ADDI(sp, sp, 64 * 8);
     as.JALR(ra);
 
-    prelude_info.end_of_prelude = reinterpret_cast<u32*>(as.GetCursorPointer());
+    prelude_info.end_of_prelude = GetCursorPtr<CodePtr>();
+}
+
+void A32AddressSpace::SetGetCursorPtr(CodePtr ptr) {
```
merryhime, line 100:
```suggestion
void A32AddressSpace::SetCursorPtr(CodePtr ptr) {
```

src/dynarmic/backend/riscv64/emit_riscv64.h
```diff
@@ -18,14 +19,21 @@ class Block;
...
+    std::ptrdiff_t code_offset;
+    LinkTarget target;
+};
+
 struct EmittedBlockInfo {
-    void* entry_point;
+    CodePtr entry_point;
     size_t size;
-    tsl::robin_map<std::ptrdiff_t, LinkTarget> relocations;
+    std::vector<Relocation> relocations;
```
merryhime, line 36: Just so it's clear as I'm aware that this isn't documented: I note the purpose of the relocation system is to make code blocks movable i.e.: to allow reordering/copying/caching of code blocks in the future (this is unimplemented).

ksco: Okay, thanks for the info!

PR #778, opened 2024-01-08
backend/rv64: Initial implementation of register allocator

ksco:
Almost identical to the initial arm64 implementation (https://github.com/merryhime/dynarmic/commit/7e046357ffc98a7ee26f005e90e21600c0a2ccfc). I'm aware that something will be changed later, and part of the code is probably shareable between arm64 and rv64, but following the steps makes the process easier for me. Hope it won't cause too much trouble for the review.


src/dynarmic/backend/riscv64/emit_riscv64.cpp
```diff
@@ -5,25 +5,97 @@
...
 #include "dynarmic/ir/basic_block.h"
+#include "dynarmic/ir/microinstruction.h"
+#include "dynarmic/ir/opcodes.h"
 
 namespace Dynarmic::Backend::RV64 {
 
-EmittedBlockInfo EmitRV64(biscuit::Assembler& as, [[maybe_unused]] IR::Block block) {
+// TODO: We should really move this to biscuit.
+static void Mov64(biscuit::Assembler& as, biscuit::GPR rd, s64 imm) {
+    if (((imm << 32) >> 32) == imm) {
```
merryhime, line 25:
```suggestion
    if (mcl::bit::sign_extend<32>(imm) == imm) {
```

`#include <mcl/bit/bit_field.hpp>`

src/dynarmic/backend/riscv64/emit_riscv64.cpp
```diff
@@ -5,25 +5,97 @@
...
+        as.LI(rd, static_cast<u32>(imm));
+        return;
+    }
+
+    // For 64-bit imm, a sequence of up to 8 instructions (i.e. LUI+ADDIW+SLLI+ADDI+SLLI+ADDI+SLLI+ADDI) is emitted.
+    // In the following, imm is processed from LSB to MSB while instruction emission is performed from MSB to LSB by calling Mov64 recursively.
+    // In each recursion, the lowest 12 bits are removed from imm and the optimal shift amount is calculated.
+    // Then, the remaining part of imm is processed recursively and as.LI() get called as soon as it fits into 32 bits.
+    s64 lo12 = (imm << 52) >> 52;
```
merryhime, line 34:
```suggestion
    s64 lo12 = s64 lo12 = mcl::bit::sign_extend<12>(imm);
```

merryhime (changes requested):

PR #782, opened 2024-01-23
backend/rv64: Bump biscuit and implements a few more features

ksco:
The biscuit subtree is re-added due to the previous wrong commit https://github.com/merryhime/dynarmic/commit/561f4189801e1576a030c7c7b240ec291709f5c7 was not recognizable by `git subtree`; added the initial implementation of terminals along with other minor features.

### DIFF
--- a/externals/biscuit/CMakeLists.txt
+++ b/externals/biscuit/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.15)
-project(biscuit VERSION 0.9.1)
+project(biscuit VERSION 0.14.0)
 
 include(CTest)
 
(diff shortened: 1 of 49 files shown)


merryhime:
Rewrote history to fix subtree issue.

lioncash:
You may want to update to the latest tag/master of biscuit, since I finally got around to getting rid of the wonky 32-bit/64-bit split on some instructions (you can just use `LI` and the normal versions of shifts to do 64-bit operations now, as opposed to LI64, SLLI64, etc).

By default the assembler will assume an RV64 target to not break backwards compatibility beyond just needing to remove "32" and "64" suffixes from instruction names. You can also override the targeting if you want to make the assembler assume RV32, or you can also make it assume RV128 (if targeting a spec with no hardware at all suits your fancy).

ksco:
> You may want to update to the latest tag/master of biscuit

Thank you, updated.

PR #785, opened 2024-01-28
Change Config to make fastmem_pointer of zero valid.

exyorha:
This changes Dynarmic::A32/A64::Config to store fastmem_pointer in a std::optional<uintptr_t>, allowing the user to pass a zero base address for the guest memory, which can be used to effectively implement a shared address space between the host and the guest.

Some additional notes:

- While an admittedly minor one, this is a breaking API change.
- I've changed both of the x64 and A64 backends, but I didn't compile nor run the A64 one, since I don't have neither a suitable build environment set up, nor anything to run it on. If you wish, I can set up something to run the tests on, but that will take some time.
- There's some space for further optimization: in the specific case when fastmem_pointer is set as zero, the allocation and the load of the fastmem base register could be omitted, and the memory accesses themselves could be emitted without the use of a base register. This could be implemented as a separate change to the backend code without further API breakage.


merryhime:
Looks fine for v7, which hopefully I'll start working on soonish.

PR #787, opened 2024-01-29
cmake: prefer system oaknut

abouvier:
Try to find the system installed `oaknut` library first (thanks to https://github.com/merryhime/oaknut/pull/15).

Also replace `QUIET` in `find_package` by the desired search mode for better information message:

- `QUIET`: no message
- `CONFIG`/`MODULE`: one line information message
- nothing: annoying multi-line information message


merryhime (approved):

abouvier:
Something was missing, because `oaknut` is used both for arm64 and for tests, regardless of the architecture.

merryhime:
> Is it normal that fuzzy tests almost always fails for dynarmic? :p

Looks like something changed in the latest xbyak. Will investigate.

abouvier:
> Looks like something changed in the latest xbyak. Will investigate.

Tests randomly failed months ago in past PRs too.

merryhime:
Random intermittent failures are normal (they're normally crashes within unicorn).
Persistent failures with xbyak exceptions like this are not, and appear to be new with xbyak 7. Reverting to 6.

PR #789, opened 2024-02-02
ir: Implement FPMulNeg

zmt00:
This should saves an instruction in both backends (and a memory access + potential cache miss on x86_64). I first considered naming the IR instruction `FPMulSub` to match the actual instruction name(s), but that seemed incorrect for an instruction that is clearly performing negation, multiplication, and addition. 

The rounding correction case has been updated to account for the need to negate an argument, but I left the NaN correction code unaltered as it does not seem to check for the sign of any of its arguments.

This is more in depth than my previous PRs - please let me know if there's anything I've missed.


zmt00:
Force-pushed to appease MSVC.

merryhime:
FPMulSub is a fine name. The operation is equivalent to opa - (op1 * op2).
I would rather use existing terminology from the architecture if possible.

Also consider: FPNegatedMulAdd would refer to a different concept (i.e. `-opa - (op1 * op2)`).

> I left the NaN correction code unaltered as it does not seem to check for the sign of any of its arguments.

You will need to negate the appropriate argument during NaN handling otherwise an incorrect NaN will be produced.

zmt00:
Thanks for the feedback - I'll close this PR for now and reopen it when I have a chance to correct it,

PR #790, opened 2024-02-04
emit_x64_vector: AVX512+GNFI implementation of EmitVectorLogicalVShift8

Wunkolo:


Wunkolo:
The AVX512{VL+BW}-requirement is purely for masking out-of-index values. Gonna look into a quicker non-AVX512 way to do this.

src/dynarmic/backend/x64/emit_x64_vector.cpp
```diff
@@ -1803,6 +1803,28 @@ void EmitX64::EmitVectorLogicalShiftRight64(EmitContext& ctx, IR::Inst* inst) {
 }
 
 void EmitX64::EmitVectorLogicalVShift8(EmitContext& ctx, IR::Inst* inst) {
+    if (code.HasHostFeature(HostFeature::AVX512_Ortho | HostFeature::AVX512BW | HostFeature::GFNI)) {
+        auto args = ctx.reg_alloc.GetArgumentInfo(inst);
+        const Xbyak::Xmm result = ctx.reg_alloc.UseScratchXmm(args[0]);
+        const Xbyak::Xmm left_shift = ctx.reg_alloc.UseScratchXmm(args[1]);
```
merryhime, line 1809:
```suggestion
        const Xbyak::Xmm left_shift = ctx.reg_alloc.UseXmm(args[1]);
```

merryhime:
I don't think this implementation is correct. It also doesn't handle negative shift values.

Wunkolo:
Should handle negative-shifts properly now.

src/dynarmic/backend/x64/emit_x64_vector.cpp
```diff
@@ -1803,6 +1803,45 @@ void EmitX64::EmitVectorLogicalShiftRight64(EmitContext& ctx, IR::Inst* inst) {
...
+        const Xbyak::Xmm tmp = ctx.reg_alloc.ScratchXmm();
+
+        const Xbyak::Opmask negative_mask = k1;
+        code.pxor(tmp, tmp);
+        code.vpcmpb(negative_mask, left_shift, tmp, CmpInt::LessThan);
+
+        // Reverse bits of negative-shifts
+        code.vgf2p8affineqb(result | negative_mask, result, code.BConst<64>(xword_b, 0x8040201008040201), 0);
```
merryhime, line 1817: `xmm0` is guaranteed to be a free scratch register.

```suggestion
        code.vmovaps(xmm0, code.BConst<64>(xword_b, 0x8040201008040201));
        code.vgf2p8affineqb(result | negative_mask, result, xmm0, 0);
```

src/dynarmic/backend/x64/emit_x64_vector.cpp
```diff
@@ -1803,6 +1803,45 @@ void EmitX64::EmitVectorLogicalShiftRight64(EmitContext& ctx, IR::Inst* inst) {
...
+        code.pand(result, tmp);
+
+        // n << 0 == n * 1 | n << 1 == n * 2 | n << 2 == n * 4 | etc
+        code.movdqa(tmp, code.Const(xword, 0x80'40'20'10'08'04'02'01, 0));
+        code.pshufb(tmp, left_shift);
+
+        code.gf2p8mulb(result, tmp);
+
+        // Un-reverse bits of negative-shifts
+        code.vgf2p8affineqb(result | negative_mask, result, code.BConst<64>(xword_b, 0x8040201008040201), 0);
```
merryhime, line 1839:
```suggestion
        code.vgf2p8affineqb(result | negative_mask, result, xmm0, 0);
```

src/dynarmic/backend/x64/emit_x64_vector.cpp
```diff
@@ -1803,6 +1803,45 @@ void EmitX64::EmitVectorLogicalShiftRight64(EmitContext& ctx, IR::Inst* inst) {
...
+
+        // gf2p8mulb's "x8 + x4 + x3 + x + 1"-polynomial-reduction only applies
+        // when the multiplication overflows. Masking away any bits that would have
+        // overflowed turns the polynomial-multiplication into regular modulo-multiplication
+        code.movdqa(tmp, code.Const(xword, 0x01'03'07'0f'1f'3f'7f'ff, 0));
+        code.vpshufb(tmp | valid_index | T_z, tmp, left_shift);
+        code.pand(result, tmp);
+
+        // n << 0 == n * 1 | n << 1 == n * 2 | n << 2 == n * 4 | etc
+        code.movdqa(tmp, code.Const(xword, 0x80'40'20'10'08'04'02'01, 0));
```
merryhime, line 1833:
```suggestion
        code.pxor(tmp, tmp);
        code.movss(tmp, xmm0);
```

merryhime (commented):
Just thinking a little about memory traffic.

src/dynarmic/backend/x64/emit_x64_vector.cpp
Wunkolo: `movsd` worked in this case since the constant is 64-bits (It looks like it repeats but the nibbles are different)

FernandoS27:
I'm not sure if we should keep pushing avx512, considering it's dead and AVX10 will replace it.

lioncash:
Fam, there's consumer hardware available right *now* that can execute AVX-512, it's fine. They're not mutually exclusive as far as support capability in dynarmic goes.

merryhime:
@FernandoS27 AVX512 support in dynarmic is basically equivalent to future AVX10 support, since subset of the AVX512 instruction set we use is within the AVX10 subset.

merryhime (approved):
Looks good. Probably the most elegant way this could've been done within available constraints.

PR #791, opened 2024-02-04
emit_x64_vector: GNFI implementation of EmitVectorCountLeadingZeros8

Wunkolo:


merryhime:
I'm happy for new tests to be written with oaknut if desired.

PR #792, opened 2024-02-07
ir: Implement FPMulSub

zmt00:
I set up unicorn to fuzz the floating point instruction implementations locally, so I'm considerably more confident in their correctness now. This should also follow ARM naming schemes for IR and related structures more closely. Please let me know if there's anything else that needs to be addressed.


merryhime (approved):

merryhime:
I feel the number of branches in NaN handling can be reduced but it's likely not worth it (requires code duplication).
Also AVX-512 implementation of NaN handling would be much simpler, but that's for future work.

PR #795, opened 2024-02-12
emit_x64_vector: Opimize VectorSignedSaturatedAbs

zmt00:
On most remotely recent processors, this should be two cycles to get the final result for sizes 8 through 32, and only a few more for 64 bits.


zmt00:
I'm unsure of why the AArch64 test for this is failing - I haven't modified any code there, and the tests pass locally without the optimization commit. Unfortunately I don't have a dev-ready machine to investigate this with.

merryhime:
> I'm unsure of why the AArch64 test for this is failing - I haven't modified any code there, and the tests pass locally without the optimization commit. Unfortunately I don't have a dev-ready ARM machine to investigate this with.

Looking into it.

merryhime:
See [4ae4750](https://github.com/merryhime/dynarmic/commit/4ae4750b5a960871c10325e0864fd8646d53f2ef).

zmt00:
Thanks! Anything else I need to do on this PR?

merryhime (approved):

PR #796, opened 2024-02-16
emit_x64_vector: Refactor pre-SSE4.1 min/max instruction replacements

zmt00:
This consolidates several cases of duplicated implementations of `p{min,max}{u,s}d` in particular. 
`EmitVectorPairedMinMaxLower8` was rearranged to be otherwise consistent with declaration order, and had its host feature requirement at call sites for unsigned values reduced from SSE 4.1 to SSSE3 - in an earlier draft of the initial PR I believe I was using `insertps`, and I didn't lower the requirement after removing it.


merryhime:
Note to self: This PR also deduplicates paired-lower fallbacks.

PR #797, opened 2024-02-18
emit_x64_vector: Implement AVX2 UnsignedRoundingShiftLeft{32,64}

zmt00:
This implementation could be expanded to support 16-bit integers with the `vps{l,r}lvw` instructions in AVX512BW+VL, but my usual dev machine doesn't have support. AVX2 is the minimum requirement for the other variable-element bit shift instructions.

(Although `xmm0` isn't required to be the mask operand for the AVX blend instructions, it did end up being convenient to do so in this particular case.)


zmt00:
Force-pushed to appease MSVC.

Wunkolo:
> but my usual dev machine doesn't have support

You can use Intel SDE to verify an implementation if your host machine does not support a particular instruction:
https://www.intel.com/content/www/us/en/developer/articles/tool/software-development-emulator.html

For example, to simulate your new unit test running on an Icelake-level machine you can use a command like this(on linux):
```
% sde64 -icl -- ./tests/dynarmic_tests "*URSHL"
```

zmt00:
Thanks, the intel simulator is interesting, I'll look into it in a future PR.

I've identified the failure case - this implementation doesn't handle left shifts of the same bit width as the register. Working on a fix now.

src/dynarmic/backend/x64/emit_x64_vector.cpp
```diff
@@ -3761,6 +3770,53 @@ static void RoundingShiftLeft(VectorArray<T>& out, const VectorArray<T>& lhs, co
...
+    // ARM documentation describes it as:
+    // res = (a + (1 << (b - 1))) >> b
+    // here, it is more convenient to implement the operation as:
+    // tmp = a >> (b - 1)
+    // res = (tmp + 1) >> 1
+    // to add 1 to the result if the last bit to be shifted off is 1
+    const Xbyak::Xmm right_shift = ctx.reg_alloc.ScratchXmm();
+    ICODE(vpadd)(right_shift, shift_amount, xmm0);
+    ICODE(vpsrlv)(right_shift, a, right_shift);
+    ICODE(vpsub)(right_shift, right_shift, xmm0);
```
merryhime, line 3805: edge-case: this results in an overflow if b = 1 and a = 0x8000'0000

merryhime: an implementation that probably avoids this follows (untested):

```cpp
ICODE(vpsrlv)(right_shift, a, shift_amount);
ICODE(vpadd)(right_shift, shift_amount, xmm0);
ICODE(vpsrlv)(shift_amount, a, shift_amount);
ICODE(vand)(shift_amount, shift_amount, code.BConst<esize>(1)); // or consider materializing 1 via ICODE(vpabs)(xmm0, xmm0);
ICODE(vaddd)(right_shift, right_shift, shift_amount);
```

zmt00: Switched to an algorithm that avoids adding the shifted-off bit until after the final shift, described in the new comment.

src/dynarmic/backend/x64/emit_x64_vector.cpp
```diff
@@ -3761,6 +3770,53 @@ static void RoundingShiftLeft(VectorArray<T>& out, const VectorArray<T>& lhs, co
...
+    code.vpand(shift_amount, shift_amount, code.BConst<esize>(xword, 0xFF));
+
+    // if b is positive, do a normal left shift
+    const Xbyak::Xmm left_shift = ctx.reg_alloc.ScratchXmm();
+    ICODE(vpsllv)(left_shift, a, shift_amount);
+
+    // if b is negative, compute the rounding right shift
+    // ARM documentation describes it as:
+    // res = (a + (1 << (b - 1))) >> b
+    // here, it is more convenient to implement the operation as:
```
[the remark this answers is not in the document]
zmt00: Adjusted documentation.

zmt00:
Thanks for the review - I've updated the right shift to avoid overflow by not adding the rounding correction bit until after the shift.

src/dynarmic/backend/x64/emit_x64_vector.cpp
```diff
@@ -3761,6 +3770,57 @@ static void RoundingShiftLeft(VectorArray<T>& out, const VectorArray<T>& lhs, co
...
+    code.vpand(right_shift, right_shift, xmm0);
+    // compute standard right shift
+    ICODE(vpsrlv)(xmm0, a, shift_amount);
+    // combine results
+    ICODE(vpadd)(right_shift, xmm0, right_shift);
+
+    // blend based on the sign bit of the lowest byte of each element of b
+    ICODE(vpsll)(xmm0, b, u8(esize - 8));
+    if constexpr (esize == 32) {
+        code.vblendvps(left_shift, left_shift, right_shift, xmm0);
```
merryhime, line 3815: Just remembered: On intel processessors, annoyingly, blendvps is cheaper than vblendvps. I note that blendvps has xmm0 as an implicit argument.

zmt00: Oh so it is... That's... curious of Intel. At least I already essentially wrote this in the SSE form, being destructive to the first source operand and with the blend mask in xmm0. Pushed an adjustment.

zmt00:
Force-pushed to appease odd microarchitecture decisions from Intel (and a brief formatting error).

merryhime (approved):

PR #798, opened 2024-02-19
backend/rv64: Implement more opcodes

ksco:
Adds minimal implementation for test case `arm: Unintended modification in SetCFlag` to work.


externals/mcl/include/mcl/assert.hpp
```diff
@@ -24,6 +24,7 @@ template<typename... Ts>
 }  // namespace mcl::detail
 
 #define UNREACHABLE() ASSERT_FALSE("Unreachable code!")
+#define UNIMPLEMENTED() ASSERT_FALSE("Unimplemented at {}:{} ", __FILE__, __LINE__);
```
merryhime, line 27: This change is occuring in a separate project and should be it's own commit so that it can be pushed upstream.

merryhime:
Don't worry about the mcl change, will update myself when I next update mcl

ksco:
> Don't worry about the mcl change, will update myself when I next update mcl

Thank you, when you do that, could you also rebase feature/rv64 to the latest?

merryhime:
Decided to push to master.

